PR #117, opened 2018-11-07
116-Clean-icon-management

### LINKED ISSUE #116: Clean icon management
ComposablePresenter>>icon: is a really bad selector name because it takes a symbol and not a form as a parameter. 

I propose to deprecate it in favor of #iconNamed:

I propose to also add ComposablePresenter>>iconNamed: directly there to be cleaner (Since Stephane wants to reduce the number of users of Object>>iconNamed:)

jecisc:
Deprecate ComposablePresenter>>#icon: to use #iconNamed:

Fixes #116


jecisc:
@estebanlm Is that too late to backport to Pharo 7 since there was the rc1?

Issue #128, opened 2018-11-07
Tabs do not need to use #asWidget


jecisc:
This is useless because the receiver is always an object and not an adaptor.

jecisc:
Fixed in https://github.com/pharo-spec/Spec/pull/130

PR #130, opened 2018-11-07
129-TabPresenter-is-too-corelated-to-Morph

### LINKED ISSUE #129: TabPresenter is too corelated to Morph
It has a method #retrievingBlock to get the content of a tab but this is a copy of the Morphic API. 

We should check the APIs of the other framework to get something more common.

jecisc:
Tab presenter and Tab example do not need to call #asWidget.

Fixes #129


jecisc:
In fact it was issue 128... :(
https://github.com/pharo-spec/Spec/issues/128

PR #141, opened 2018-11-27
132-refactor-DialogWindowPresenter

### LINKED ISSUE #132: refactor DialogWindowPresenter
DialogWindowPresenter is too restrictive: 

- it is able to display just "Ok"/"Cancel" options.
- it assumes the actionBar will be added as another component (which is not always the case). 

I propose to refactor it to : 

- accept different buttons (and by default offer Ok/Cancel, to keep backward compatibility)
- do not use OkCancelToolbar
- translate the platform specific code to the adapter.

estebanlm:
Closes #132

### DIFF
--- a/src/Spec-MorphicAdapters/SpecDialogWindow.class.st
+++ b/src/Spec-MorphicAdapters/SpecDialogWindow.class.st
@@ -27,14 +27,13 @@ SpecDialogWindow >> cancelled [
 { #category : #actions }
 SpecDialogWindow >> close [
 
-	super close.
-	self toolbar cancel
+	super close
 ]
 
 { #category : #actions }
 SpecDialogWindow >> escapePressed [
 
-	self toolbar triggerCancelAction
+	self model triggerCancelAction
 ]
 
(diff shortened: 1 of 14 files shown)


jecisc:
With this PR this code is broken:

```Smalltalk
DatePresenter example
```

jecisc:
Before the focusOrder of the presenter contained a OkCancelPresenter but now it contains a nil.

src/Spec-MorphicAdapters/SpecDialogWindow.class.st
```diff
@@ -99,7 +98,7 @@ SpecDialogWindow >> newMainPanel [
 	| panel |
 	panel := self newDialogPanel.
 	self newContentMorph ifNotNil: [:m | panel 	addMorphBack: m ].
```
jecisc:
```suggestion
	self newContentMorph ifNotNil: [:m | panel addMorphBack: m ].
```

jecisc:
I tested and it seems fine now.

PR #166, opened 2018-12-28
165-Backport-PluggableTextMorph-changes-from-pharo-and-rewrite-the-input-number

### LINKED ISSUE #165: Backport PluggableTextMorph changes from pharo and rewrite the input number
A PR was integrated in Pharo to replace the usages of PluggableTextMorph by Rubric. 

The input text used it, so we should update it. 

Also, the input number is totally broken with this change and should be rewritten.

jecisc:
Use rubric instead of PluggableTextMorph for input text and number.

Fixes #165


jecisc:
I'll close for now because we need others changes in Pharo 8 before.

PR #176, opened 2019-01-15
170-Merge-ListPresenter-and-FastTablePresenter

jecisc:
Use FastTable to build ListPresenters.


guillep:
The tests I've introduced earlier are failing. I'll merge it and fix them.
What do you think if we make FastTablePresenter just an empty (deprecated) subclass of ListPresenter or TablePresenter? The idea is to not break the code of people using it directly.

jecisc:
It was a work in progress for some features and in the end we wanted to do something else. That’s why I did not merged it. (I did not add [IGNORE] because there was only Esteban and me at the beginning of this project).

Issue #197, opened 2019-01-25
Reimplement TabManagerPresenter


estebanlm:
This is more like a new presenter (I will do a `NotebookPresenter`).
The reason is simple: `TabManagerMorph `is plain bad.

estebanlm:
The problem with this is that LazyTabGroupMorph is not good for our usage since it does not works fine with morphs instead plain strings as tabs (and we want to be able to add icons and buttons there).

So, I'm working on cleaning a bit that part before committing the new presenter.

estebanlm:
related to #162

Issue #162, opened 2018-12-26
Morphic tabs content should be lazily loaded


jecisc:
Currently, in the morphic binding of the TabPresenter, when we open the tab presenter, all tabs content are initialized. 

If you have something that is long to display, you need to not set any content and set it when you select the tab. 

This is not good, by default the content should only be computed when we select the tab. Not at opening

jecisc:
@estebanlm is working on that.

estebanlm:
yup, and almost finished :)

Issue #191, opened 2019-01-22
Unify api events #when*:


jecisc:
Sometimes we have `whenSomething:` and sometimes we have `whenSomethingDo:`.

We should only keep one IMO.

jecisc:
I would take `whenSomethingDo:` because it is the API of `Model`.

PR #208, opened 2019-01-29
add SpecPanedLayout and NotebookPresenter

estebanlm:
and deprecating TabManagerPresenter.


estebanlm:
sorry for the huge PR.
adding functionality that needs new functionality causes that :P

estebanlm:
the full PanedLayout is a hack to properly introduce initial positions for the panels there there.
but well... it works :P

src/Spec-Core/NotebookPresenter.class.st
```diff
@@ -0,0 +1,118 @@
...
+]
+
+{ #category : #private }
+NotebookPresenter >> updatePageContent: aPage [
+
+	self changed: #updatePageContent: with: { aPage }
+]
+
+{ #category : #'api-events' }
+NotebookPresenter >> whenSelectedPageChanged: aBlock [
```
jecisc: @estebanlm See issue https://github.com/pharo-spec/Spec/issues/191

I would like your opinion on it. If we go for the #when*Do: form, I'll update, this PR.

estebanlm: In general I prefer without,  but I'm open to one or the other... but we need to use just one :)
(in fact I would have preferred other prefix/semantics (style `onChangedSelection:`), but is too late for that ;)

jecisc: Both are fine with me. 

I proposed the #when*Do: way because we have #whenChangedDo: in Model and NewValueHolder inherits from Model.

guillep:
Looks ok to me overall, with the deprecation and all. Too big for a pull request and to review it in detail...

estebanlm:
*shrugs* PRs can be done adding functionality. 
Now, obviously reviewing them is hard. As it is now, is just a way to say "This is ready to integration" :P

estebanlm:
now, I don't know why it failed :P

estebanlm:
also, 
Fixes #162  
Fixes #197

Issue #211, opened 2019-01-29
Implement single selection in TablePresenter


guillep:
Child of issue #195

guillep:
Related: pharo-project/pharo#2405

guillep:
Fixed in #214

PR #214, opened 2019-01-29
211-Implement-single-selection-in-TablePresenter

guillep:
Fix issue #211. Implement Single selection in TablePresenter with tests

### DIFF
--- a/src/Spec-Core/TableBasePresenter.class.st
+++ b/src/Spec-Core/TableBasePresenter.class.st
@@ -55,6 +55,7 @@ TableBasePresenter >> initialize [
 	showColumnHeadersHolder := true asValueHolder.
 	columns := #().
 	isResizableHolder := false asValueHolder.
+	selectionHolder := SelectionValueHolder new.
 	
 	showColumnHeadersHolder whenChangedDo: [ self showColumnHeadersChanged ]
 ]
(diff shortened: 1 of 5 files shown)


src/Spec-Core/TableBasePresenter.class.st
```diff
@@ -75,6 +76,41 @@ TableBasePresenter >> isShowingColumnHeaders [
...
+]
+
+{ #category : #api }
+TableBasePresenter >> selectedItem [
+	
+	^ selectionHolder selection value
+]
+
+{ #category : #api }
+TableBasePresenter >> setSelectedIndex: anIndex [
```
estebanlm: I don't like the name `setBlabla:`
in fact, I was going to schedule a task for deprecate all setter alla java names in Spec (there are a lot).
Would it be possible to change it?

guillep: Yes, I can change it, but still I think the life cycle would be much faster if we are not forced to make a PR for every little change we do...

Also, we should discuss conventions not only about that but also about events, rename everything "d'un coup" and make tests. Most of those things can be done fairly fast and with automatic deprecations.

jecisc: At the beginning we went for PR like that we can do:
- Review
- A changelog (because with the amount of commit I only check PR to do the changelog, I don't have the time to go through all commits)
- A migration guide

If we commit directly then we have no review and it's impossible to keep track of changes. And since Spec is already in used in a lot of projects, people will just abandon Spec if we do releases without changelog and migration guide. (Some people already told me that if it's too hard to migrate to the new Spec they will keep their project in Pharo 7 or stop using Spec)

guillep: That's not necessary true.
- If we do a review of every single commit I do, we will spend 50% of our time doing reviews. And I'll be blocked 50% of my time waiting for an issue to be integrated.
- Making direct pushes does not mean I cannot create issues. Issues can be used for the changes log.
- The same for the migration guide.

We can keep some backwards compatibility, deprecations with automatic rewriting and forward compatibility packages to help people migrate.

But all those do not go against of commiting directly. It's just extra bureaucracy.

guillep: I think for this particular case this could be a normal accessor (`#selectedIndex` and `#selectedIndex:`).

@demarey I've read that book like 10 years ago already and you will see that this is not the case KentBeck is talking about

Now, just to stop discussing for nothing, we have talked and direct pushes will be ok :)

guillep: Latest 2 commits should solve this conversation

estebanlm: @demarey as Stef says, common accessors/setters (and this is just a common accessor, despite the fact of being a value holder) does not need/wants "set". 
Also, in general I do not like set either (even if Kent Beck recommended it): I prefer no mutators but actions. 
IMO (but this is overkill for now, since everything is made like that), `selectedIndex:` is also bad (but passable) because it kind of unveils is going to affect a property and no more, while `selectIndex:` would be more correct because it describes the action to execute which will finish changing the selectedIndex property.
Anyway, just a reflection to consider :)

demarey: @estebanlm ok. I thought you wanted to remove all methods starting by set. At some point, it would be nice to sit all together and agree on code style convention for Pharo. We then can write a document  from it and share it.

PR #279, opened 2019-02-19
Fixing selection holder value change in #setSelectionInterval: of TextPresenter

### LINKED ISSUE #278: Selection holder in TextPresenter does not trigger when setSelectionInterval: is called
See method #setSelectionInterval: in TextPresenter class : it calls rawValue: method on the selection holder, which does not reports the change. The #value: method should be used instead.

StevenCostiou:
Using #value: instead of #rawValue: in #setSelecionInterval: of TextPresenter. It correctly triggers the value change announcement and calls the #whenChangedDo: behavior.
Fixes #278


jecisc:
@StevenCostiou I think you told me that this change might be wrong it the end. Do I remember well? 

Do we close the PR or will you fix the tests?

StevenCostiou:
Sorry I missed the comment :/ It sould be closed (I'm doing it following this comment). 
The change is wrong but the issue is real ;)

PR #288, opened 2019-02-21
better handing of list refreshing items

estebanlm:
it ensures the transmission happens


src/Spec-Core/SpecInterpreter.class.st
```diff
@@ -11,6 +11,9 @@ Class {
 		'presenter',
 		'selector'
 	],
+	#classInstVars : [
```
jecisc: This variable should not be here.

src/Spec-Core/SpecPragmaCollector.class.st
```diff
@@ -9,6 +9,9 @@ Class {
 	#instVars : [
 		'behavior'
 	],
+	#classInstVars : [
```
jecisc: This variable should not be here.

src/Spec-Core/ComposablePresenter.class.st
```diff
@@ -1191,13 +1189,14 @@ ComposablePresenter >> show [
 { #category : #accessing }
 ComposablePresenter >> spec [
 
-	^ spec
+	^ adapter
 ]
 
 { #category : #accessing }
-ComposablePresenter >> spec: aSpec [
+ComposablePresenter >> spec: anAdapter [
```
jecisc: We should deprecate those methods with a rewrite rule but I can do it later.

src/Spec-Inspector/EyeInspector.class.st
```diff
@@ -11,6 +11,9 @@ Class {
 		'list',
 		'description'
 	],
+	#classInstVars : [
+		'useAutoRefresh'
```
jecisc: This variable should not be here.

estebanlm: uh? 
I didn't add them.

jecisc: I removed them in a previous commit. Maybe Iceberg got mixup?

src/Spec-Core/SpecPragmaCollector.class.st
jecisc: This one should still be removed.

src/Spec-Core/SpecInterpreter.class.st
estebanlm: which one?

src/Spec-Core/SpecPragmaCollector.class.st
estebanlm: still... I didn't touch any of this.
if they are there, is not my fault :)

PR #345, opened 2019-02-28
344-Remove-users-of-FastTablePresenter-to-use-ListPresenter

### LINKED ISSUE #344: Remove users of FastTablePresenter to use ListPresenter

jecisc:
Replace usage of FastTablePresenter to use ListPresenter.

There is still one user but it uses a table and not a list...

Fixes #344


jecisc:
There was a failure but I cannot reproduce locally. I'll relaunch the tests.

PR #351, opened 2019-03-01
191-Unify-api-events-when

### LINKED ISSUE #191: Unify api events #when*:

jecisc:
Unify #when*: and #when*Do:

+ Be sure they are in the right protocol (api-events)

Fixes #191


jecisc:
Failure is unrelated.

PR #367, opened 2019-03-04
introduce-application

estebanlm:
lots of broken stuff... but is almost usable :)


src/Spec-Core/Application.class.st
```diff
@@ -0,0 +1,100 @@
+"
+I'm a Spec application entry point. 
+In Spec, each application needs to have a definition that concentrates its flow and resources.
+
+"
+Class {
+	#name : #Application,
```
jecisc: Wouldn't it be better to call it SpecApplicaiton while we do not have modules in Pharo?

tesonep: To me it is the same. With that idea we have to prefix everything with spec?

jecisc: In a system without modules/namespaces, yes I think. I know multiple projects with a concept of "Application". If everyone do not care about the problem, nobody will be able to load anything in Pharo in no time.

estebanlm: Mmm, but we are not "nobody": We are doing the base framework :)
I renamed it from `SpecApplication` to just `Application` because I think if we use a prefix, we need it for everything (Then, we should name things `SpecPresenter`, etc.). Spec does not use prefixes... regardless of this is/was a good choice at start, it is like that. 
Let's discuss then the general use of prefixes. Do we want a prefix for Spec? Will we apply that prefix to everything? Otherwise, is just not coherent.

estebanlm: (Note that I'm not against, I just want to be coherent)

jecisc: I would personally prefer all of Spec to be prefixed.

PR #373, opened 2019-03-05
334-harvest-changes-from-Pharo [WIP]

### LINKED ISSUE #334: harvest changes from Pharo
During all these months, things in Spec may have changed (recategorizations...). We should harvest them, but since lots of things have changed, the merge may be painful. So we should:
 - merge latest changes into the **FIRST** commit in spec
 - merge that commit into master

That would delegate all the merge pain to git

guillep:
Fix #334
Harvesting changes made required for Pharo integration

 - fixing cyclic dependencies (examples should be in example package)
 - fixing baseline problems
 - fixing categorizations for release tests

Still we need to fix the baseline hack


src/BaselineOfSpec/BaselineOfSpec.class.st
```diff
@@ -7,21 +7,6 @@ Class {
 	#category : #BaselineOfSpec
 }
 
-{ #category : #'pre/post load' }
-BaselineOfSpec >> addHackToLoadSpec [
```
jecisc: Removing the hacks will work for now because there is only few changes between Pharo8 and Spec version, but if at some point we create a change that will make a package dirty *in the middle* of an update, it will crash once again. 

I don't think it should be removed just like that without any solution. The case I described before can just be achieved by moving a class to another package. So it is not uncommon.

guillep: That's why I did not integrate it blindly and made a PR for this one :).
Before integrating this one I want to have a green CI.

Already loading in Pharo7 should give an scenario like the one you say, no?

jecisc:
I re-did the PR with the latest changes and I'll close this one.

PR #384, opened 2019-03-08
381-SpecDemo-window-is-small

### LINKED ISSUE #381: SpecDemo window is small
The extent of the SpecDemo window (and probably other applications with custom window size) is small

jecisc:
Fix extent of demos.

Fixes #381


jecisc:
Same failures as in master.

PR #386, opened 2019-03-08
347-addModelInwithSpecLayout-methods-should-be-renamed

### LINKED ISSUE #347: addModelIn:withSpecLayout: methods should be renamed
methods like MorphicWindowAdapter>>#addModelIn:withSpecLayout: should be renamed to do not mention "model"

jecisc:
Fix name consistency to remove #model and use #presenter.

Fixes #347


jecisc:
Some of the failures are related.

jecisc:
Now it's the same amount of failing test than master.

PR #388, opened 2019-03-08
387-Remove-baseline-hack

### LINKED ISSUE #387: Remove baseline hack
There was a hack in Spec baseline to update it. But now loading is atomic and it fix the problem. We can now remove the hack I think.

jecisc:
Remove baseline hack.

Fixes #387


jecisc:
Same amount of failing test as in master.

PR #409, opened 2019-03-21
408-Changing-the-width-of-a-SpecBoxLayout-horizontal-changes-the-height-instead

### LINKED ISSUE #408: Changing the width of a SpecBoxLayout horizontal changes the height instead
I have this spec:

```Smalltalk
defaultSpec
	^ SpecBoxLayout newVertical
		add: #modeDropbox withConstraints: [ :constraints | constraints height: 30 ];
		add:
			(SpecBoxLayout newHorizontal
				add: #list withConstraints: [ :constraints | constraints width: 100 ];
				add: #table;
				yourself);
		yourself
```

The problem is that when I open this presenter, the list still takes 50% of the width, but its height is of 100px.

jecisc:
Fix width constraint of box layout.

Fixes #408


jecisc:
I would like a review of @estebanlm because I had to make the constraint aware of their layout to make this work. Is it fine with you or do you think there is a better way to correct this problem?

jecisc:
ping

PR #418, opened 2019-03-25
417-List-presenter-does-not-take-all-height-in-horizontal-paned-layout

### LINKED ISSUE #417: List presenter does not take all height in horizontal paned layout
When using a Paned layout with a list inside, the list will not take the full height.
Screenshot 2019-03-25 at 15 47 05

jecisc:
List should fill verticaly by default. 

Fixes #417


jecisc:
Same failing test as Master.

PR #422, opened 2019-03-26
cleanup/layout-and-interpreter

### LINKED ISSUE #416: Pass on dynamic presenters
We should be able to add/change layout elements dynamically from a presenter

guillep:
Fix #416

Several cleanups to layouting and interpreter.
 - extracting legacy (Array based) layout into another object to cleanup presenter implementation
 - composable presenters may have layout instances and talk to them
 - layouts can be updated at runtime


guillep:
Well, in the same PR I fix the CI for Pharo 8. I'm checking the one of Pharo7 and I cannot get it failing...

PR #444, opened 2019-04-01
443-Deprecate-useless-presenters

### LINKED ISSUE #443: Deprecate useless presenters
In Spec we have some really simple presenters that should not be there IMO. 

For example, there is a presenter to associate a button and a drop box. In the end, it duplicates all the API of both of them which is bad because when we change the API of a widget, then they are not in sync. And they are so simple that the Spec user could do it himself in his presenters. 

Targeted:
- AbstractTwoButtons
- DropListButton
- InputTextDropList
- LabelledDropListButton
- LabelledInputTextDropList
- PickListPresenter

jecisc:
Deprecate classes with really small values.

Fixes #443


jecisc:
CI passed + nobody complained. I merge.

PR #458, opened 2019-04-10
391-GridLayout-is-not-able-to-rebuild-itself

### LINKED ISSUE #391: GridLayout is not able to rebuild itself
When you need to on the fly to replace one presenter with another one, you need to do:
```
	self needRebuild: false.
	self buildWithSpec.
```
However, this code does not work as soon as you use the SpecGridLayout (and maybe others too). See the attached code

[TstPresenter.zip](https://github.com/pharo-spec/Spec/files/2945982/TstPresenter.zip)

jecisc:
Add a way to rebuild executables layout.

This PR is for the Morphic backend.

Fixes #391

### DIFF
--- a/src/Spec-Examples/SpecDemo.class.st
+++ b/src/Spec-Examples/SpecDemo.class.st
@@ -17,23 +17,15 @@ Class {
 
 { #category : #specs }
 SpecDemo class >> defaultSpec [
-
-	<spec: #default>
-
-	| fontWidth listWidth |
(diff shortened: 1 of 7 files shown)


estebanlm:
does this has a test?

jecisc:
I added SpecExecutableLayoutTest >> testCanDynamicallyRebuildPresenter and SpecExecutableLayoutTest >> testRebuildPresenterBeforeOpeningWorks

PR #464, opened 2019-04-12
463-Cannot-have-condition-in-menus

### LINKED ISSUE #463: Cannot have condition in menus
Before menus on lists where built when the user did a right click on an element of the list. This allowed the user to use conditions to add/remove/disable/change the title of menus.

Now the menu is built at the opening of a window. We should update that.

jecisc:
Add conditions to menu.

Fixes #463


jecisc:
Now Pharo 8 also timeout... Don't know where the problem comes from :(

jecisc:
Same failure than master.

PR #470, opened 2019-04-16
466-Cannot-addremove-page-to-Notebook

### LINKED ISSUE #226: Add Notebook demo
Esteban added a Notebook presenter. We should also add a demo in SpecDemo.

### LINKED ISSUE #466: Cannot add/remove page to Notebook
[Archive.zip](https://github.com/pharo-spec/Spec/files/3084079/Archive.zip)

Here is the beginning of what I did. But it needs a manual rebuild. We should not have to do that.

### LINKED ISSUE #468: Notebook pages should be able to open with non default layout
Currently, a page only opens presenters with their default layout. We should be able to specify a non default layout also.

jecisc:
Work on Notebook presenter.

- Add possibility to use a non default spec
- Add/Remove page
- Add presenter and backend tests
- Improve API 
- Fix programmatic selection 

Fixes #466
Fixes #468
Fixes #226

### DIFF
--- a/src/Spec-BackendTests/NotebookAdapterTest.class.st
+++ b/src/Spec-BackendTests/NotebookAdapterTest.class.st
@@ -8,3 +8,47 @@ Class {
 NotebookAdapterTest >> classToTest [
 	^ NotebookPresenter
 ]
+
+{ #category : #running }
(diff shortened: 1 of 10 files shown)


jecisc:
Same failures than master

PR #472, opened 2019-04-17
471-Pharo-7-CI-is-timeouting

### LINKED ISSUE #471: Pharo 7 CI is timeouting

jecisc:
In Pharo 7 we need to recompile the code after loading Spec 2.

Fixes #471


jecisc:
Failing tests of Pharo 8 are the same as master and Pharo 7 now is running even if it has failing tests.

PR #493, opened 2019-04-19
334-harvest-changes-from-Pharo

### LINKED ISSUE #165: Backport PluggableTextMorph changes from pharo and rewrite the input number

### LINKED ISSUE #167: Remove usage of RubPluggableTextFieldMorph
Before Spec used a PluggableTextMorph for the input number and text.

Now it is replaced by RubPluggableTextMorph, but this morph will be removed later and is just here for compatibility.

We should remove its usage in Spec.

### LINKED ISSUE #334: harvest changes from Pharo

jecisc:
Harvest Pharo changes.

Fixes #334
Fixes #167 
Fixes #165


jecisc:
This is weird because the failing test pass on my computer and I don't see what could have mess with it in this PR.

PR #505, opened 2019-05-02
add the contextMenu to the TreeTablePresenter

570gina:
I tried to add some codes to the TreeTablePresenter and got successful to use it on my project.


570gina:
I omitted the register event.

PR #508, opened 2019-05-09
507-add-contextMenu--to-TreeTablePresenter-

CafeKrem:
add contextMenu to TreeTablePresenter and test but it's just copy and paste


jecisc:
I would like to show the feature in the Spec demo but it's a first step. I checked and it works!

PR #527, opened 2019-05-20
Add "sortingBlock" to the DropListPresenter

570gina:
I need to use the sort feature so I add some methods about sortBlock in the DropListPresenter
(Refer from AbstractListPresenter)


jecisc:
Thank you!

Issue #499, opened 2019-04-25
Multiple selection doesn't remove the path that has been deselected


570gina:
I deselected the selected item, but the selection of TreeMultipleSelectionMode is still including the item.

570gina:
for example,
I selected item1, item2, item3 and I got the selection including these 3 elements.
Then I deselected item1 but I still got three instead of two.

PR #549, opened 2019-05-27
499 multiple selection doesnt remove the path that has been deselected

570gina:
I won't get an item that has been deselected:)

I modified some code to solve the issue #499


jecisc:
I tested and it works well. 

I'll merge as soon as the CI is green.

PR #638, opened 2019-07-05
637-add-a-UI-to-select-Method

CafeKrem:
add new UI for choose a method


jecisc:
This PR contains the removal of stuff that should not be here.

It happens because Spec is not in a clean state at loading, I don't know yet why.

PR #652, opened 2019-07-15
Additionnals features required for migrating Iceberg to Spec 2

Inao0:
add "iconWidth" "LabelWidth"
add "add:height:" "add:width" for adding spec layout with one constraint more easily.
add "item:" to button bar to be able to directly set all the buttons.
add a compatibility package to make spec2 work with commander.
add a "getShortcutTipForMenuItem" to allow CommandActivator to generate their shortcut as a String.


jecisc:
As we discussed IRL, I'm not sure the Commander compatibility package should be here since Commander is loaded after Spec. 

I'll review the PR when it will be updated.

src/Spec2-Layout/SpExecutableLayout.class.st
```diff
@@ -15,6 +15,18 @@ SpExecutableLayout >> adapterName [
 	^ self subclassResponsibility
 ]
 
+{ #category : #'as yet unclassified' }
+SpExecutableLayout >> add: aPresenter height: anInteger [
```
jecisc: Those methods should probably be on SpBoxLayout only since the other layout constraints don't have #height: and #width:

PR #655, opened 2019-07-17
ifNil and variants usage

gcotelli:
Use `ifNil:`,  `ifNotNil:` and `ifNil:ifNotNil:` instead of `isNil` and `ifTrue/ifFalse` combinations


jecisc:
Thank you!

PR #656, opened 2019-07-17
Making Spec2 load in Pharo7

tesonep:
I have fixed the dependencies to check that Spec loads in Pharo 7


src/Spec2-Examples/SpSearchableList.extension.st
```diff
@@ -1,10 +0,0 @@
-Extension { #name : #SpSearchableList }
```
jecisc: Is this removal wanted?

jecisc: Oh, it was just repackaged.

PR #738, opened 2019-08-17
676-Observable-slot-should-not-launch-a-Value-changed-event-when-the-value-is-the-same-than-the-previous-one

### LINKED ISSUE #676: Observable slot should not launch a Value changed event when the value is the same than the previous one
For example, if we clear multiple time the content of a list, only the first one should be announced. 

Possible way:

```Smalltalk
SpValueHolder>>value: anObject
	"Handle circular references as explained in the class comment"
	lock ifTrue: [ ^ self ].
	
	value = anObject ifTrue: [ ^ self ].

	lock := true.
	
	[ | oldValue |
	oldValue := value.
	value := anObject.
	self valueChanged: oldValue ]
		ensure: [ lock := false ]
```

jecisc:
Do not update observable value holder if the value did not changed.

Fixes #676

### DIFF
--- a/src/Spec2-Core/SpButtonPresenter.class.st
+++ b/src/Spec2-Core/SpButtonPresenter.class.st
@@ -213,7 +213,7 @@ SpButtonPresenter >> performAction [
 	self action value.
 
 	" Here I set a dummy value just to make the holder raise an event "
-	actionPerformed := nil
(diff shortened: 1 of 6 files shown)


jecisc:
This PR broke more test than failing tests in master.

estebanlm:
this PR is not valid anymore, and needs to be re-done :(

PR #747, opened 2019-08-29
Adding the ability to defer a UI rendering operation to the backend.

Inao0:
Required to update the UI properly in Iceberg (Spec 2 version) PullRequestsBrowser.
The aim is to able to fork a process that will fetch the needed data to then ask the presenter to properly updates it's state in the right thread.


jecisc:
Thank you!

PR #751, opened 2019-09-04
748-Use-correct-Spec2-prefix-also-for-traits

### LINKED ISSUE #748: Use correct Spec2 prefix also for traits
Spec2 should be prefixed with "Sp" - this should be true also for the traits. 
The order is first the prefix "Sp" and then the "T" like Traits as the prefix acts like a namespace.

So the following traits need to be renamed
```
TSpDynamicPresenter -> SpTDynamicPresenter
TSpHaveWrappingScrollBars -> SpTHaveWrappingScrollBars
TSpObservable -> SpTObservable
TSpMorphicTableDataSourceCommons -> SpTMorphicTableDataSourceCommons
```

jecisc:
Use SpT prefix instead of TSp for traits.

Fixes #748

### DIFF
--- a/src/Spec2-Adapters-Morphic/SpAbstractMorphicAdapter.class.st
+++ b/src/Spec2-Adapters-Morphic/SpAbstractMorphicAdapter.class.st
@@ -209,6 +209,13 @@ SpAbstractMorphicAdapter >> extent [
 	^ self widget extent
 ]
 
+{ #category : #'api-focus' }
+SpAbstractMorphicAdapter >> gtInspectorPreviewIn: composite [
+	<gtInspectorPresentationOrder: 30>
+	
+	self widgetDo: [ :w | w gtInspectorMorphIn: composite ]
+]
+
 { #category : #protocol }
 SpAbstractMorphicAdapter >> hRigid [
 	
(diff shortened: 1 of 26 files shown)


jecisc:
I don't know why but this does not work in Pharo 7... :(

astares:
Why not?

jecisc:
```
MessageNotUnderstood: receiver of "asTraitComposition" is nil

UndefinedObject(Object)>>doesNotUnderstand: #asTraitComposition

MCClassDefinition>>traitCompositionCompiled

[ :builder | 

builder

	superclass: superClass;

	name: name;

	layoutClass: (ObjectLayout layoutForType: type);

	slots: self instanceVariables;

	sharedVariables: self classVariables;

	sharedPools: self sharedPoolsString;

	classSlots: self classInstanceVariables;

	traitComposition: self traitCompositionCompiled;

	classTraitComposition: self classTraitCompositionCompiled;

	comment: comment stamp: commentStamp;
```

I don't know yet the origin of the problem.

jecisc:
We plan to remove TSpObservableSlot from Spec to push it in Pharo directly. We'll rename it when doing that. I'll wait for this to be integrated into Pharo before updating this PR.

estebanlm:
I think this has been fixed upstream

PR #779, opened 2019-09-13
harvest-pharo-changes

jecisc:
Spec list presenter scrolling.


jecisc:
This is weird because it does not load in Pharo 7 but I cannot reproduce it locally.

PR #803, opened 2019-10-10
Use-Pharo-Observable-Slots

jecisc:
Migrate Spec to ObservableSlots from Pharo


estebanlm:
this PR now is invalid, I will schedule a task to do it again later ;)

PR #804, opened 2019-10-16
feature/make-composablelist-a-list

estebanlm:
add selection (single/multiple) to component list


estebanlm:
there is this random problem on P7 :(

Issue #818, opened 2019-12-12
Improve SpEditableList


demarey:
Add a layout to use editable list without up,down, top, bottom buttons.
More configurable (items display, label).
Add tests

PR #819, opened 2019-12-12
818-Improve-SpEditableList

demarey:
Improve EditableList. Fixes issue #818


src/Spec2-Core/SpEditableList.class.st
```diff
@@ -23,45 +23,69 @@ Class {
...
-		'okBlock'
+		'okBlock',
+		'label'
 	],
 	#category : #'Spec2-Core-Widgets'
 }
 
+{ #category : #accessing }
+SpEditableList class >> buttonWidth [
+	^ 26
```
estebanlm: nope. 
Button width can't be hardcoded. 
Please use an style from SpStyle :)

src/Spec2-Core/SpEditableList.class.st
```diff
@@ -23,45 +23,69 @@ Class {
...
 		add:
 			(SpBoxLayout newHorizontal
 				add: #list;
 				add:
 					(SpBoxLayout newVertical
-							add: #topButton;
-							add: #upButton;
-							add: #downButton;
-							add: #bottomButton;
+							add: '';
```
estebanlm: nope. 
Do not use fixed sizes!

src/Spec2-Core/SpEditableList.class.st
```diff
@@ -23,45 +23,69 @@ Class {
...
+	^ 40
+]
+
+{ #category : #spec }
+SpEditableList class >> withoutOrderingBar [
+	^ SpBoxLayout newVertical
+		add:
+			(SpBoxLayout newHorizontal
+					add: #label;
+					add: #addButton width: self buttonWidth;
```
estebanlm: do not use fixed sizes!

estebanlm:
So, the correct way to affect button sizes is through styles. 
You can add a "small button style" in SpStyle, something like: 

```
...
	.button [  
		Geometry { #width: 100 }
                .small {
		    Geometry { #width: 30 }
               }
	],
...
```

then in your presenter: 

```Smalltalk
...
upButton := self newButton addStyle: 'small'.
...       
```

estebanlm (changes requested):

estebanlm:
This is not cool : 

```Smalltalk
(SpBoxLayout newVertical
							add: '';
							add: #topButton height: self orderingButtonHeight;
							add: #upButton height: self orderingButtonHeight;
							add: #downButton height: self orderingButtonHeight;
							add: #bottomButton height: self orderingButtonHeight;
							add: ' ';
							yourself)
```

Something like this you have to do it this way: 

```Smalltalk
(SpBoxLayout newVertical
    add: #topButton expand false;
    add: #upButton expand false;
    add: #downButton expand false;
    add: #bottomButton expand false;
   yourself)
```

Do not use an empty label to create a space!
Do not fix the heights!

PR #821, opened 2019-12-13
820-allow-droplist-in-toolbar

### LINKED ISSUE #820: allow droplist in toolbar
currently, only buttons are accepted in a toolbar.
I have a use case where I would like to use a drop list in the toolbar. Looks like other apps use it also (e.g. intellij idea)

demarey:
Add support for droplist in toolbar. Fixes #820

### DIFF
--- a/src/Spec2-Adapters-Morphic/SpDropListMorph.class.st
+++ b/src/Spec2-Adapters-Morphic/SpDropListMorph.class.st
@@ -10,6 +10,11 @@ Class {
 	#category : #'Spec2-Adapters-Morphic-Support'
 }
 
+{ #category : #'as yet unclassified' }
+SpDropListMorph >> configureWith: displayModel item: itemPresenter [ 
+	displayModel configureDropList: self item: itemPresenter
+]
+
 { #category : #private }
 SpDropListMorph >> currentIcon [
 
--- a/src/Spec2-Adapters-Morphic/SpDropListMorph.class.st
+++ b/src/Spec2-Adapters-Morphic/SpDropListMorph.class.st
@@ -10,7 +10,7 @@ Class {
 	#category : #'Spec2-Adapters-Morphic-Support'
 }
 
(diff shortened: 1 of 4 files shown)


estebanlm:
Damn it. In fact this is wrong, toolbars do accept drop lists. 
In case of need, you can use a menu button to supply that.
I think I will need to revert this @demarey :(

astares:
@estebanlm: Can you explain. Why shoud a toolbar not allow for drop lists?

estebanlm:
because toolbars do not do that :)
toolbars have certain features all elements needs to fulfil: 
- each element can be formatted to a mode: icons, text, icons + text. 
- toolbars can be displayed in horizontal or vertical layout (even if we support just horizontal at the moment).
- etc?

then, you can not put arbitrary elements in a toolbar (I guess a ribbon would be that?). In Gtk you can put: buttons, toggle buttons, menu buttons and that's all. Cocoa allows a similar set of elements (plus a search bar). 
So, we cannot support that and pretend to support other backends than morphic.

But... nothing forbids you to combine toolbar + drop lists (or anything) into a panel than later you can put wherever you want. But that dropdown cannot be *part* of the toolbar. Look this example: 

Screenshot 2020-03-18 at 17 11 13

this is in fact a toolbar + a panel with a search presenter, make like this: 

```Smalltalk
defaultSpec

	^ SpBoxLayout newVertical 
		add: (SpBoxLayout newHorizontal
				spacing: 5; 
				add: #toolbar;
				addLast: (SpBoxLayout newVertical 
					add: #searchField 
						expand: false 
						fill: false
						padding: 5;
					yourself) 
					expand: false
					fill: false
					padding: 5;
				yourself)
			expand: false; 
		add: (SpPanedLayout newHorizontal
			position: 30 percent;
			add: #indexPanel;
			add: #listPanel;
			yourself);
		yourself
```

PR #824, opened 2019-12-30
Feature/dynamic layouts

guillep:
Recover my changes from before parents leave ^^.
This PR adds support for dynamically modifying the layout of a presenter.

Typically, a presenter should initialize itself like:

```smalltalk
initializePresenters
	
	layout := SpBoxLayout newHorizontal
```

And then add/remove elements to the layout programatically instead of using the **declarative** class-side method.


estebanlm:
lets of tests failing :P 
also, I need to adapt this to gtk ;)

estebanlm:
ok, I will merge this ;)

PR #834, opened 2020-02-10
Enhance EditablePresenter

demarey:
Use an editable collection.
Allow selection by index.


estebanlm:
conflicting now

demarey:
conflict should now be solved

PR #837, opened 2020-02-13
stringsLocalization

pavel-krivanek:
labes localization


pavel-krivanek:
Basically, it assumes that the strings in the spec application may be something more than just plain strings. Something that can produce strings depending on the current presenter (or by default, application) locale.
The goal is to be able to switch the language without the need for recreating of everything

src/Spec2-Adapters-Morphic/SpAbstractMorphicAdapter.class.st
```diff
@@ -376,7 +376,9 @@ SpAbstractMorphicAdapter >> rightClicked [
 { #category : #protocol }
 SpAbstractMorphicAdapter >> setBalloonText: aString [
 
-	self widget ifNotNil: [ :w | w setBalloonText: aString ]
+	self widget ifNotNil: [ :w | 
+		aString ifNotNil: [ :theString |
```
estebanlm: it may happen that string is nil (and this will remove the balloon).

src/Spec2-Core/SpAbstractPresenter.class.st
```diff
@@ -371,6 +377,12 @@ SpAbstractPresenter >> update: aParameter [
 	self changed: aParameter
 ]
 
+{ #category : #'private utils' }
+SpAbstractPresenter >> updateSlot: aSymbol [
```
estebanlm: I prefer something like "forceUpdateSlot" here.

src/Spec2-Core/SpMenuPresenter.class.st
```diff
@@ -28,7 +28,7 @@ Class {
 	#superclass : #SpAbstractWidgetPresenter,
 	#instVars : [
 		'#groups',
-		'#title',
+		'#title => SpObservableSlot',
```
estebanlm: what is this title for? (I know you just made it observable, but for me this was in "marked for removal")

src/Spec2-Core/String.extension.st
```diff
@@ -1,9 +1,32 @@
 Extension { #name : #String }
 
+{ #category : #'*Spec2-Core' }
+String >> accentuateCharacter: aCharacter [
```
estebanlm: if you are answering a copy (or a new instance), I prefer something like `withAccentuatedCharacter:` as name.

src/Spec2-Core/SpMenuPresenter.class.st
pavel-krivanek: Titles for menus do not make much sense but because it is there and I wanted to support translation of it, I needed to change it to the observable slot to allow changes propagation

src/Spec2-Core/String.extension.st
pavel-krivanek: renamed

src/Spec2-Core/SpAbstractPresenter.class.st
pavel-krivanek: renamed

PR #838, opened 2020-02-20
List-Scrolling-Fix

StevenCostiou:
Additional  scrolling updates when selection is changed in lists.
(Error in branch name, should be list and not table).


estebanlm:
I'm not a fan on the source formatting of this change. 
It makes harder to read it and follow what's happening. I think in the future I will be more strict about the ways to structure source code while contributing (every big project has his style guidelines, why we shouldn't)
The change is fine so I will accept it (also I'm ranting at you when I should rant in general :P).  
but please, please, please, think on format better your code. 
Do it for the future generations. 
Do it for the future you.

PR #846, opened 2020-03-20
Cleanup/world hand references

guillep:
Harvest changes in Pharo.
Cleanups in World, ActiveHand, ActiveWorld and related.


demarey:
Does it mean that Spec will not work on Pharo 8?

PR #860, opened 2020-04-02
world-presenter-enhancements

demarey:
remove useless method

### DIFF
--- a/src/Spec2-Adapters-Morphic/SpMorphicWorldAdapter.class.st
+++ b/src/Spec2-Adapters-Morphic/SpMorphicWorldAdapter.class.st
@@ -25,3 +25,10 @@ SpMorphicWorldAdapter >> buildWidget [
 SpMorphicWorldAdapter >> centered [
 	"Do nothing. I will fill the full world"
 ]
+
+{ #category : #accessing }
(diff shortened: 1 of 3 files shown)


src/Spec2-Adapters-Morphic/SpMorphicWorldAdapter.class.st
```diff
@@ -19,9 +19,26 @@ SpMorphicWorldAdapter >> addPresenterIn: widgetToBuild withSpecLayout: aSpec [
 SpMorphicWorldAdapter >> buildWidget [
 
 	^ self presenter currentWorld 
+		borderColor: self theme backgroundColor;
+		borderWidth: 4;
```
estebanlm, line 23: this should be done through a style, I think?

estebanlm: still, I can accept it because this is too morphic specific. Tell me if it is too hard.

demarey: yes, really morphic specific. Also, I do not see any use of border in the style nor how to use colors from the theme. Let's go fast for this time. We will review that with gtk backend

PR #861, opened 2020-04-06
add methods and tests to interact on  toolbar toggle button state change

demarey:


demarey:
forget this PR. I will do a new one

PR #868, opened 2020-04-10
fix scrolling

### LINKED ISSUE #866: List selection has strange behaviour with scrolling
You can see an example of this strange behaviour here: https://github.com/pharo-project/pharo-launcher/issues/444
This behavior was introduced in https://github.com/pharo-spec/Spec/pull/838.

Here is a small script you can use to reproduce:
```smalltalk
presenter := SpTablePresenter new.
presenter
	items: (1 to: 100) asArray;
	openWithSpec. "stop execution here"

presenter selectIndex: 36. "execute and look at the presenter. Should look ok"
presenter selectIndex: 38. "execute and look at the presenter. Here, scrolling should not have happened"
```

I discussed with @guillep of this issue. Here is a short summary:

- API was designed to decouple scrolling from selection
- The PR:
  - introduces coupling of scrolling and selection
  - uses private methods (ex: https://github.com/pharo-spec/Spec/blob/master/src/Spec2-Adapters-Morphic/SpAbstractMorphicListAdapter.class.st#L37)
- We should use ```presenter verticalAlignment desiredVisibleRow: index```instead

Now a design question is: **should selection scroll?**
- for single selection, why not?
- what happens for multiple selection when all selected elements do not fit in the visible elements?

@guillep proposes to add this method to the presenter:
```smalltalk
selectIndex: anInteger scrollToSelection: scrollToSelection
    self selectIndex: anInteger
    scrollToSelection ifTrue: [ 
        self verticalAlignment desiredVisibleRow: anInteger ].
```

@estebanlm @StevenCostiou @tesonep your opinion?

demarey:
Fixes #866

### DIFF
--- a/src/Spec2-Adapters-Morphic/SpAbstractMorphicListAdapter.class.st
+++ b/src/Spec2-Adapters-Morphic/SpAbstractMorphicListAdapter.class.st
@@ -28,13 +28,6 @@ SpAbstractMorphicListAdapter >> configureScrolling [
 		onAnnouncement: FTScrollingChanged
 		send: #scrollingChanged
 		to: self.
-		
-	self flag: 'This is a fix to correctly update the scrolling when updating selection. I am not certain this is the right way to do it.'.	
-	self presenter selection
-		whenChangedDo: [ widget
-				scrollToIndex: self presenter verticalAlignment desiredVisibleRow.
-			widget selectedIndex > widget lastVisibleRowIndex
-				ifTrue: [ widget moveShowIndexTo: widget selectedIndex ] ]
 ]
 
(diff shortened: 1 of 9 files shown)


demarey:
Failing tests are not related. They are already failing in the master branch ...

guillep (approved):
It looks ok for me.

This PR does also include scrolling support for tables besides lists (+test), does it? just making it explicit ^^

demarey:
yes, it also adds scrolling to trees. Also, trees now uses SpVerticalAlignement just as lists.

estebanlm (approved):

PR #870, opened 2020-04-14
Fix scrolling

demarey:
fix scrolling


demarey:
unwanted commits in the PR => reject

PR #895, opened 2020-06-12
fix #516 Fix memory leak in SpApplication not releasing closed morphic windows

### LINKED ISSUE #516: Memory leak - the default application remembers all windows
The default SpecApplication still remembers all windows when they were closed. That, of course, produces a huge memory leak.

demarey:


src/Spec2-Core/SpWindowPresenter.class.st
```diff
@@ -194,13 +194,13 @@ SpWindowPresenter >> centeredRelativeToTopWindow [
 	self centeredRelativeTo: self application topWindow
 ]
 
-{ #category : #api }
+{ #category : #updating }
```
estebanlm, line 197: why `#close` is on updating?

demarey: looks like it is unattended. I do not remember updating the protocol. We can set it back to #api

Issue #806, opened 2019-10-26
Notebook: can't add a closable page


tinchodias:
The adapter seems to support it, though.

tinchodias:
I come back to this report after downloading the latest P9 and loading Spec2's master (*) and trying `SpNotebookPage>>#beCloseable`. 

1. Found no demos or tests for this method (I installed default group of Spec2, don't know if I should install another group).
2. I modified this method: `SpDemoNotebookPresenter>>#objectClassPage`

```Smalltalk
objectClassPage
	^ (SpNotebookPage
		   title: 'Object class'
		   icon: (self iconNamed: #nautilusIcon)
		   provider: [ MessageBrowser new messages: Object methods ])
		  beCloseable;
		  yourself
```

3. Open that example with `SpDemoNotebookPresenter new openWithSpec`.

<img width="909" alt="Screen Shot 2020-07-20 at 12 40 06" src="https://user-images.githubusercontent.com/3044265/87963078-2ac1e500-ca86-11ea-9e7f-4b5dc087dbc4.png">

while it originally looked like:
<img width="733" alt="Screen Shot 2020-07-20 at 12 43 08" src="https://user-images.githubusercontent.com/3044265/87963416-ad4aa480-ca86-11ea-90eb-1b0b9727b5e0.png">

So:
- The closable tab show no close button 
- All tabs are broken



(*) My install script in P9 was:
```smalltalk
    Metacello new
        githubUser: 'pharo-spec' project: 'Spec' commitish: 'master' path: 'src';
        baseline: 'Spec2';
        onConflict: [ :e | e useIncoming ];
        onUpgrade: [ :e | e useIncoming ];
        ignoreImage;
        load
```

PR #904, opened 2020-07-29
Notebook: can't add a closable page

tinchodias:
Fix to #806.
Not sure if it's the best solution, but to me it looks evident that nobody was using `beCloseable` and this makes it work.


estebanlm:
thanks!

PR #905, opened 2020-07-31
Missing super

StephanEggermont:
Allow the closing of modal dialogs 

See https://github.com/pharo-project/pharo/issues/7032 and  https://github.com/pharo-project/pharo/pull/7033


estebanlm:
thanks!

Issue #933, opened 2020-09-03
TextPresenter context menu not fully overridden; accumulates multiple copies of group


martinmcclure:
When I override the context menu of a CodePresenter (because I want only my own options, not all the default ones, like this:

```
codePresenter := (self instantiate: SpCodePresenter)
                         contextMenu: self paneMenu;
                         overrideContextMenu: true;
                         yourself.
```
The menu I expect (in this little example) is this:
![image](https://user-images.githubusercontent.com/2240523/92071904-d4bead80-ed64-11ea-99be-664b105224cf.png)
But the context menu I get is this:
![image](https://user-images.githubusercontent.com/2240523/92071958-f324a900-ed64-11ea-946f-10fecc5f94ac.png)
... the first time. Each time I open the context menu, another copy of the unwanted group is added, so the menu gets longer and longer. For instance, the second time I open the menu it looks like this:
![image](https://user-images.githubusercontent.com/2240523/92072037-236c4780-ed65-11ea-9314-8f30c8a64879.png)

estebanlm:
this was fixed in 826e01de4b27b28219571595a6bcbc34a8d35951
Now, I changed the idiom to override context menu from `overrideContextMenu: aBoolean` to `overridingContextMenu` (because it matches a bit more the spirit of Spec2).

You have an example at `SpCodePresenter class>>#exampleOverridingContextMenu` (and to just remove the edition command in `SpTextPresenter class>>#exampleWithoutEditionContextMenu`

PR #936, opened 2020-09-08
Reverse sense of #allowingContextMenu

martinmcclure:
It looks like #allowingContextMenu, introduced as part of the fix for issue #933, had its sense accidentally backward.


estebanlm:
ouch!

PR #939, opened 2020-09-10
Added SpListWithFilterPresenter and SpListWithFilterPresenterTest

### LINKED ISSUE #938: [Feature Request] add a SpListWithFilterPresenter
As the filtering functionality is removed of SpListPresenter, I think that will be very useful to have a filtering list presenter.
The name can be: SpListWithFilterPresenter. This can be achieved using a high level component that has an SpListPresenter and SpTextInputFieldPresenter.
I have been working on this and I think that I found a solution but the code must be reviewed.

jordanmontt:
Fixes #938

### DIFF
--- a/src/Spec2-Core/SpListWithFilterPresenter.class.st
+++ b/src/Spec2-Core/SpListWithFilterPresenter.class.st
@@ -1,0 +1,126 @@
+"
+I am a presenter that has a list and a text input for filtering the elements on the list.
+The items on the list must be set using the method `self items:`, not using the accessor `self list` and then passing the message `items:`. This is because an internal variable is needed in order to recover the original list items when the filter text is erased.
+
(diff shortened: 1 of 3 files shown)


jordanmontt:
I have been thinking and believe that SpListWithFilterPresenter should inherit from SpListPresenter. This because if you want to modify several behaviors of the list, you will have to:
1) call several times the list accessor. i.e. listWithFilter list whenSelectionChangedDo: - listWithFilter list headerTitle: - listWithFilter list display: - listWithFilter list icons: -  listWithFilter list sortingBlock: ... etc
2) or to have all list api methods implemented on SpListWithFilterPresenter
But I don't know... what do you think?

jordanmontt:
SpListWithFilter should not inherit from SpListPresenter. Because it is not a good design and since we are adding a SpTextInputFieldPresenter it will not be possible. I have made a commit right now adding more tests and making the code more clear. I think that the presenter is finished.

PR #946, opened 2020-10-01
Add an accessor for tree selectionMode

pavel-krivanek:


src/Spec2-Core/SpTreeTablePresenter.class.st
```diff
@@ -322,6 +322,12 @@ SpTreeTablePresenter >> selection [
 	^ selectionMode value
 ]
 
+{ #category : #accessing }
```
estebanlm, line 325: this is now category "api"
... and it needs a method comment (you can use the same as in `SpAbstractListPresenter>>#selecitonMode)` :)

estebanlm (changes requested):

pavel-krivanek:
redone

PR #947, opened 2020-10-02
Fix for SpSliderPresenter value returns values in the range 0..1 and not min..max fixes #943

### LINKED ISSUE #943: SpSliderPresenter #value returns values in the range 0..1 and not #min..#max
**Fix:**

 Changing value to scaled seems to work(last line)

majella67:
Spec Fix


estebanlm:
please update your branch with latest changes, otherwise I will not be able to merge

estebanlm:
... or if it was a small change (as I think), you can submit a new PR and close this... is maybe easier.

PR #950, opened 2020-10-06
jordanmontt/938-add-SpListWithFilterPresenter

Ducasse:
filtering list presenter from sebastian revisited and ready for integration


estebanlm:
manually applied the changes

PR #955, opened 2020-10-19
Fix visibility of workspace variables in debugIt expressions

dionisiydk:
Fixes integration between SpCodePresenter and morphic widget to properly setup a requestor for related compilation operations:
- debugIt command
- code highlighting

It fixes the visibility of workspace variables in debugIt expressions and allows further improvement to have transparent temps.

The fix requires extra PR for NewTools to have an effect: https://github.com/pharo-spec/NewTools/pull/122.

### DIFF
--- a/src/Spec2-Code-Commands/SpCodeDebugItCommand.class.st
+++ b/src/Spec2-Code-Commands/SpCodeDebugItCommand.class.st
@@ -35,7 +35,7 @@ SpCodeDebugItCommand >> compile: aStream for: anObject in: evalContext [
 		source: aStream;
 		class: methodClass;
 		context: evalContext;
-		requestor: nil;
+		requestor: context; "it should enable a visibility of current tool variables in new debugger"
 		noPattern: true;
 		failBlock: [ ^ nil ];
 		compile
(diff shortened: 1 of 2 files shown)


dionisiydk:
I did not found existing tests for modified classes. If you will point me where to add them I will do

Ducasse:
Thanks denis!

estebanlm:
this one is rather for @StevenCostiou :)

MarcusDenker (approved):
I tested it: with this we see the workspace variable in the debugger and we can print it.

Highlighting works, too.

MarcusDenker:
It interestingly works *without* the change for NewTools: https://github.com/pharo-spec/NewTools/pull/122

dionisiydk:
> It interestingly works _without_ the change for NewTools: [pharo-spec/NewTools#122](https://github.com/pharo-spec/NewTools/pull/122)

Hm.  I do not see that it works without this PR. 
(but integration is safe without it)

MarcusDenker:
very odd. I added the changes by hand to the latest Pharo9 and then I see a var defined in the Playground in the debugger.

my test is: 

add var by "test := 1", doit.

Then I do "test halt" --> it is shown as a variable in the debugger, I can evaluate "test" and get 1.

dionisiydk:
yes, it's really strange. I did exactly what you said with latest Pharo 9.

dionisiydk:
I suspect that you are using GTDebugger. It may be due to saved settings.
Old debugger always supports workspace vars. No changes were needed there.
Am I right?

PR #959, opened 2020-10-28
Fixed bug in SpFilteringListPresenter

jordanmontt:
It was a bug: When changing the items of a spFilteringList that already had items, the filter functionality stops working.
For example, if you do `SpFilteringListPresenter new items: anItems`. Typing something in the filter box and then setting the new items `filteringList items: newItems`. The filter functionality stops working. The bug was fixed.

Also, the instance variable name of filterInput was changed to filterInputBlock.

And finally, the test for the filterListItems; method now is more detailed.


estebanlm:
thanks !

PR #1009, opened 2021-03-02
fixing-moving-buttons-in-toolbar

tesonep:
The border style of the toolbar button should not have style different than the one when it is in normal state.

Having additional border makes it bigger and it moves, producing a flick.
The SpToolbarMorph should be rigid vertically and spaceFill horizontally to honor the values set has height by the SpToolbarDisplayMode
The extent when the SpToolbarDisplayModeIconAndLabel should be square.

Fix https://github.com/pharo-vcs/iceberg/issues/1431

### DIFF
--- a/src/Spec2-Core/SpToolbarDisplayModeIconAndLabel.class.st
+++ b/src/Spec2-Core/SpToolbarDisplayModeIconAndLabel.class.st
@@ -17,7 +17,7 @@ SpToolbarDisplayModeIconAndLabel >> configureButton: aButton item: aToolbarItem 
 
 { #category : #accessing }
 SpToolbarDisplayModeIconAndLabel >> extent [
-	^ 45@40
+	^ 45@45
 ]
 
 { #category : #accessing }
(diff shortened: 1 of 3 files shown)


tinchodias:
looks good

src/Spec2-Core/SpToolbarDisplayModeIconAndLabel.class.st
```diff
@@ -17,7 +17,7 @@ SpToolbarDisplayModeIconAndLabel >> configureButton: aButton item: aToolbarItem
 
 { #category : #accessing }
 SpToolbarDisplayModeIconAndLabel >> extent [
-	^ 45@40
+	^ 45@45
```
estebanlm, line 20: I do not like this one. 40 is already very big. Why is this needed?

tesonep: To squared. if not we can use 40@40

estebanlm: mmm... ok ?
let´s test it.

PR #1011, opened 2021-03-03
The printing of the shortcuts should delegate in the KMShortcutPrinter.

tesonep:
Fix https://github.com/pharo-vcs/iceberg/issues/1409


estebanlm:
ah yes... that code was made before I made the KMShortcutPrinter :P

PR #1021, opened 2021-03-29
Harvest differences to spec as in build 1254

guillep:
I took the latest master branch in spec in build 1254 and it has difference, which is strange because a merge happened today? Also I don't understand, some of these seem to have been changed in the last commits of spec. So this PR seems to revert them, It's not clear to me.


estebanlm:
this one is incorrect, is removing methods that should be there

PR #1034, opened 2021-04-20
Rename SpSliderInput to SpSliderInputPresenter.

NourDjihan:
SpSliderInput is a presenter, it makes more sens if the class name ends with the suffix presenter.


estebanlm:
this is conflicting now, but change is good, I will apply the rename upstream :)

PR #1035, opened 2021-04-20
Rename SpPopoverPosition hierarchy.

NourDjihan:
The hierarchy becomes consistent if the 'Position' suffix is placed at the end of all subclasses + class names make more sens.
Example: SpPopoverBottomPosition makes more sens than SpPopoverPositionBottom

### DIFF
--- a/src/Spec2-Core/SpPopoverBottomPosition.class.st
+++ b/src/Spec2-Core/SpPopoverBottomPosition.class.st
@@ -1,0 +1,14 @@
+"
+I define a popover needs to be shown at the bottom of its owner.
+"
+Class {
+	#name : #SpPopoverBottomPosition,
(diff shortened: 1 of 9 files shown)


astares:
Please deprecate the old name then. There might be users.

NourDjihan:
Which class is not deprecated?

estebanlm:
not really, the way to use this is not accessing directly the classes but through SpPopoverPosition class accessor. 
And even that, is accessed through SpPopoverPresenter>>bePosition*  family. 

Finally, another thing is that Spec2 is not yet in stable release (not arrived to 1.0) and start deprecating things before releasing is not how we do (sometimes I can consider it, but not in general... until we are not beta anymore).

PR #1036, opened 2021-04-20
Renaming SpToolbarItemPosition hierarchy.

NourDjihan:
SpToolbarItemRightPosition and SpToolbarItemLeftPosition make more sens than SpToolbarItemPositionRight and SpToolbarItemPositionLeft respectively + coherent .


estebanlm:
even if I am considering removing this functionality (because you can have the same result by composing two toolbars), change is good :)

PR #1038, opened 2021-04-20
Rename SpTestApplicationWithLocale to SpWithLocaleTestApplication .

NourJihene:
For the coherency of the system SpWithLocaleTestApplication seems better

### DIFF
--- a/src/Spec2-Morphic-Backend-Tests/SpApplicationWithLocaleTest.class.st
+++ b/src/Spec2-Morphic-Backend-Tests/SpApplicationWithLocaleTest.class.st
@@ -7,7 +7,7 @@ Class {
 { #category : #tests }
 SpApplicationWithLocaleTest >> testLabel [
 	| app label string |
-	app := SpTestApplicationWithLocale new.
+	app := SpWithLocaleTestApplication new.
 	app locale: (Locale isoLanguage: 'en' isoCountry: 'US').
 	app useBackend: #Morphic.
 	string := SpTestLocalizedString from: 'string'.
--- a/src/Spec2-Tests/SpTestApplicationWithLocale.class.st
+++ b/src/Spec2-Tests/SpTestApplicationWithLocale.class.st
@@ -1,26 +1,35 @@
+"
+Please comment me using the following template inspired by Class Responsibility Collaborator (CRC) design:
+
+For the Class part:  State a one line summary. For example, ""I represent a paragraph of text"".
+
+For the Responsibility part: Three sentences about my main responsibilities - what I do, what I know.
+
+For the Collaborators Part: State my main collaborators and one line about how I interact with them. 
+
+Public API and Key Messages
+
+- message one   
+- message two 
+- (for bonus points) how to create instances.
+
+   One simple example is simply gorgeous.
+ 
+Internal Representation and Key Implementation Points.
+
+    Instance Variables
+	locale:		<Object>
+
+
+    Implementation Points
+"
 Class {
 	#name : #SpTestApplicationWithLocale,
-	#superclass : #SpApplication,
-	#instVars : [
-		'locale'
-	],
+	#superclass : #SpWithLocaleTestApplication,
 	#category : #'Spec2-Tests-Localization'
 }
 
-{ #category : #accessing }
-SpTestApplicationWithLocale >> initialize [ 
-
-	super initialize.
-	
-	locale := Locale isoLanguage: 'en' isoCountry: 'US'.
-]
-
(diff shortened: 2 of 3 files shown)


astares:
Sounds strange - I would either 
- keep existing name "SpTestApplicationWithLocale" or use 
- "**SpLocalizationTestApplication**" / "**SpLocalizedTestApplication**"

NourDjihan:
Hum I don't know the functionality of the class .. But, I assume  **SpLocalizationTestApplication** is when a TestApplication class is responsible for localizing something, and **SpLocalizedTestApplication** is when the TestApplication itself is localized,

However, I Think that **SpLocaleWithTestApplication** seems more close in meaning to SpTestApplicationWithLocale without really knowing the actual meaning :-)

astares:
@NourDjihan You propose a new (IMHO strange name) without actually checking / understanding the functionality of the class?

But the nice thing is that we have the source, so it is easy to find out:
1.  SpTestApplicationWithLocale inherits from SpApplication, so it is an application based on Spec framework
2.  SpApplication implements a method #locale returning the current locale (which is the locale of the underlying platform running Pharo):

```Smalltalk
locale
	^ Locale current
```
which is basically found out by a primitive (if you check more deeply in Locale class. So in France it would return "fr-FR", in germany "de-DE", ...

- SpTestApplicationWithLocale provides an additional instance variable "locale" compared to what is inherited from super class which is initialized "hard coded" for Locale en-US in SpTestApplicationWithLocale>>#initialize but can be changed with getters and setters
- the class is referenced in SpApplicationWithLocaleTest>>#testLabel (which is a unit test) testing that labels could be used with a specific locale

=> So the class **SpTestApplicationWithLocale** is a test application where one can influence the locale used for labels and other dynamically using a custom locale (and not relying on the platforms default locale).
 
If we follow your proposal "SpLocaleWithTestApplication" it would sound as if there is a single locale with a test application - which is not the case and would be wrong. The locale does NOT have the test application - because it is the other way around: it is the test application that has the locale (stored in its instance variable).

So it is either a:
 - SpTestApplicationWithLocale (current name) or
 - SpTestApplicationWithCustomLocale (even more precise) or
 - SpLocalizedTestApplication

NourDjihan:
Le mer. 21 avr. 2021 à 00:41, Astares ***@***.***> a écrit :

> @NourDjihan <https://github.com/NourDjihan> You propose a new (IMHO
> strange name) without actually checking / understanding the functionality
> of the class?
>
> But the nice thing is that we have the source, so it is easy to find out:
>
>    1. SpTestApplicationWithLocale inherits from SpApplication, so it is
>    an application based on Spec framework
>
> Indeed, I have checked the inheritance tree of the class, which is
actually the main motive of the renaming proposal,

SpApplication implements a method #locale returning the current locale
(which is the locale of the underlying platform running Pharo):

> locale
> 	^ Locale current
>
> which is basically found out by a primitive (if you check more deeply in
> Locale class. So in France it would return "fr-FR", in germany "de-DE", ...
>
>    - SpTestApplicationWithLocale provides an additional instance variable
>    "locale" compared to what is inherited from super class which is
>    initialized "hard coded" for Locale en-US in
>    SpTestApplicationWithLocale>>#initialize but can be changed with getters
>    and setters
>    - the class is referenced in SpApplicationWithLocaleTest>>#testLabel
>    (which is a unit test) testing that labels could be used with a specific
>    locale
>
> What I meant is:  My understanding of the class is limited to some lines
of code , so basically I can only assume the functionality of the class
based on my personal understanding, which might be wrong just like my
suggestion :)


>
> => So the class *SpTestApplicationWithLocale* is a test application where
> one can influence the locale used for labels and other dynamically using a
> custom locale (and not relying on the platforms default locale).
>
> If we follow your proposal "SpLocaleWithTestApplication" it would sound as
> if there is a single locale with a test application - which is not the case
> and would be wrong. The locale does NOT have the test application - because
> it is the other way around: it is the test application that has the locale
> (stored in its instance variable).
>
I appreciate making the effort to further explain the functionality of the
class

> So it is either a:
>
>    - SpTestApplicationWithLocale (current name) or
>    - SpTestApplicationWithCustomLocale (even more precise) or
>    - SpLocalizedTestApplication
>
>
Nour

>
>
> —
>
Nour

estebanlm:
In this case, I would not rename this class.
(or in any case, nor of the solutions satisfies me).
The purpose of that class is: 
- is a TestApplication, 
- that implements a a locator to be used precisely to test that functionality
So, IMO (not english speaker) SpTestApplicationWithLocale is very intention revealing.

PR #1062, opened 2021-05-05
Renaming SpRGBSliders to SpRGBSlidersPresenter

NourDjihan:
The suffix Presenter helps understand that this class is a actually a Presenter


estebanlm:
conflicting, but applying the changes upstream :)

PR #1063, opened 2021-05-05
Renaming SpMethodToolbar to SpMethodToolbarPresenter.

NourDjihan:
SpMethodToolbar is a Presenter :)


estebanlm:
applied upstream

PR #1064, opened 2021-05-05
Renaming SpStyleVariableEnvironment to SpStyleEnvironmentVariable

NourDjihan:
Since it is a base for environment variables, it is better to rentame it to SpStyleEnvironmentVariable.


NourDjihan:
I Will re-do the whole hierarchy at once in one PR to avoid conflits.

PR #1095, opened 2021-06-09
Cleanup categorization issues also in Spec/Spec2

### LINKED ISSUE #1094: Cleanup categorization issues also in Spec/Spec2
Do a pass on class and primarily method categorization also for Spec/Spec2:

- some classes are not tagged (uncategorized classes in some packages)
- some methods are uncategorized still
- some testXXX methods are in "running" but should be in category "tests" 
- the method #classToTest is sometimes in "accessing" which is fine, but also sometimes wrongly in "running"
- the method #beforeTest: should be in "running" (one of the two was in "accessing")
- the method #newLayout should be in "accessing" (some of them were in "running")
- use "private - xxxx" categories instead of "private-xxxx" (so we stay consistent with others in image)
- remove some unnecessary dots at end of method
- use category name "tests" to categorize test methods, not "test"
- apply formatting to a few methods
- method category "initialization - private" should be "private - initialization" to follow standard image and get the red icon in browser for private methods
- use "utilities" instead of "utils"
- ...

astares:
Fix #1094

### DIFF
--- a/src/Spec2-Adapters-Morphic/SpAbstractMorphicAdapter.class.st
+++ b/src/Spec2-Adapters-Morphic/SpAbstractMorphicAdapter.class.st
@@ -128,7 +128,7 @@ SpAbstractMorphicAdapter >> applyVisibility [
 			ifFalse: [ self widget hide ] ]
 ]
 
-{ #category : #'protocol-shortcuts' }
+{ #category : #'protocol - shortcuts' }
 SpAbstractMorphicAdapter >> bindKeyCombination: aShortcut toAction: aBlock [
 
 	self widgetDo: [ :w | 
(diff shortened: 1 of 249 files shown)


estebanlm:
I have to say that I like more both of the other variants (`private accessing` and `private-accessing) than the one you are using :/, but this is just personal judgement, so I really do not care about it... I guess you are unifying this kind of protocols everywhere?

astares:
Yes - the image had different patterns and variants (due to the different preference of people primarily and importing of different code). 

The unifying should:
 - try to provide a common style for the most used protocols (and try to avoid further different writings)
 - give a more unified selection in the "New protocol" dialog (as it queries the system)

image

image

To decide which one to use I checked:
 - which ones were the most common used in internal and external projects already
 - which one is more readable (especially when having several of similar kind above each other the " - " separation is like a an own categorization in the categorization and makes this more explicit) and may later possibly help in further features (like having a tree instead of a list for the method protocols)
 - also it could be useful to separate with " - " when we want to extend meta facilities later and make "private - xxx" really "private" in the sense of not callable from the outside or other

image

PR #1101, opened 2021-06-17
Unsubscribe adapters

pavel-krivanek:
add a mechanism to unsubscribe adapters that are replaced by another ones


src/Spec2-Core/SpAbstractAdapter.class.st
```diff
@@ -195,6 +199,26 @@ SpAbstractAdapter >> sendRightClickEvent [
 	self subclassResponsibility
 ]
 
+{ #category : #releasing }
+SpAbstractAdapter >> unsubscribe [
```
estebanlm, line 203: shouldn't this be in "private" ?

pavel-krivanek: I wanted something more meaningful... if `private` then `private-releasing` :-)

estebanlm: with spaces `private - releasing`. Otherwise we will force Torsten to fix it ;)

Issue #1068, opened 2021-05-12
SpTextPresenter>>text: memory leak.


kgardas:
There is a memory leak somewhere in Pharo's VM or spec implementation. It's probably SpObservableSlot related. It's easily duplicated by:
```
tt := SpTextPresenter new.
tt openWithSpec.
1 to: 1000000 do: [ :n | tt text: (n asString,' test of line.') ].
```
It's probably slot related since if I try:
```
1 to: 1000000 do: [ :n | tt acceptText: (n asString,' test of line.').
            tt notifyPropertyChanged: #text. ].
```
it shows the memleak, but if I remove notifyPropertyChanged call, then memory leak is gone.

estebanlm:
so, I have been looking at this... why do you say there is a "memory leak"?
I executed your test and observed the memory without seeing an increment of it that would show the leak... 
So now I am lost :)

astares:
@kgardas Please try in a fresh image - be aware that Spec now uses ObservableSlot.

SpObservableSlot should have no users anymore.

kgardas:
@estebanlm @astares Guys, please easy on me. :-) What I see or saw while reporting the issue was increasing RAM usage of the Pharo VM. Concretely with example above I'm able to go from 138MB RES to more than 490MB RES. The memory stays there even after code is finished and even after I execute `Smalltalk garbageCollect` -- so that's the reason I called that "memleak".

Now, the test above was done on build 1310 64bit Linux VM. It's possible that the issue is already solved. Please allow me to test on newest build and if it's solved there I may go for witch hunt and binary search the "hero" build which solved the issue. Thanks for your fast and very kind support!

kgardas:
I've obtained todays' most recent 64bit Linux build number 1484 and drum rolls, the memory leak as described is still there. I'm using exactly this code:
```
tt := SpTextPresenter new.
tt openWithSpec.
1 to: 1000000 do: [ :n | tt text: (n asString,' test of line.') ].
```
just put that into playground, selected whole and right-click and select "Do It". Wait and wait and observe memory consumption growing of this particular VM in separate terminal window in htop -- in my case.

Please allow the code to run -- here it seems to run for several minutes till it opens text presenter window -- at this time whole memory leak should be visible.

Please let me know if you are able to duplicate or not. btw, running this on up-to-date Ubuntu 20.04.x LTS if that matters. FYI: RAM in this particular case go from 106MB till 488MB where it stayed.

kgardas:
[...]

PR #1105, opened 2021-06-21
fixes leaks on ObservableSlot and Adapters

estebanlm:


estebanlm:
this should fix  #1068, but we need to verify

PR #1125, opened 2021-08-01
SpDemo: add Presenters to focusOrder, not Layout

### LINKED ISSUE #1124: SpDemo: cycling focus using <tab> causes DNU
**Bug description**
When using the <tab> key to cycle throught the widgets in the SpDemo window, the following error is generated: "Instance of SpBoxLayout did not understand #traverseInFocusOrderDo:excluding:".

**To Reproduce**
1. Open an instance of the new Spec UI Framework Demo by selecting the menu item Help > Spec2 examples
2. Press the <tab> key
3. See that a debugger window is opened, with the error message above

**Expected behavior**
Focus should cycle between the visible widgets in the Demo window

**Version information:**
Pharo-9.0.0+build.1542.sha.7864e16cc43b34e13f97099516375103a59aca00 (64 Bit) - OS: [e.g. iOS]

EwanDawson:
Fixes #1124 

This change assumes that Layouts shouldn't be added to `focusOrder`, only Presenters.

### DIFF
--- a/src/Spec2-Examples/SpDemo.class.st
+++ b/src/Spec2-Examples/SpDemo.class.st
@@ -139,7 +139,7 @@ SpDemo >> initializePresenters [
 
 	self focusOrder
 		add: list;
-		add: page
+		addAll: page presenters 
 ]
 
 { #category : #accessing }


EwanDawson:
I'm new to Spec, and I'm not sure if it makes sense to add Layouts to focusOrder or not. If we do want to allow layouts to be added to focusOrder, then we'd need to extend `SpExecutableLayout` to understand `#traverseInFocusOrderDo:excluding:`:

```smalltalk
SpExecutableLayout >> traverseInFocusOrderDo: aBlock excluding: excludes [

	self presenters do: [ :each | each traverseInFocusOrderDo: aBlock excluding: excludes ]

]
```

I've tested the above change, and it solves the issue too. Let me know if this is preferable, and I'll make that change instead.

Also worth noting that if I remove the explicit setting of focus order at the end of `SpDemo>>#initializePresenters`, that fixes the error too. This is because, if `focusOrder` is `nil`, then the result of `layoutPresenters` is used as the focus order. But then `focusOrder` is lazily initialized by its accessor, so relying on iit being `nil` does not seem to be robust.

Ducasse:
Thanks for your issue. We will handle it probably beginning of september because of holidays.

PR #1142, opened 2021-09-06
merge latest changes

estebanlm:


Ducasse:
This is strange because I do not see the changes we did for the navigator. 
I will check it.

PR #1147, opened 2021-09-16
fixes https://github.com/pharo-spec/NewTools/issues/266

MarcusDenker:
- read the value of the Global before browsing


MarcusDenker:
No idea why I can not create a PR...

MarcusDenker:
Why is this a conflict?

MarcusDenker:
I resolved the conflict, but there is something wrong that I do when doing a PR

MarcusDenker:
I merged that small change into https://github.com/pharo-spec/Spec/pull/1150 which touches the same method (as such, faster turnaround would help...)

PR #1179, opened 2021-10-19
Fix #1176: restore code from #1150

### LINKED ISSUE #1151: [Cleanup] remove pool var related code from SpCodePresenter
Somehow I can not commit are removal of these methods (changing the sender to not call them worked and was already merged)

SpCodePresenter>>#browseImplementorOfPoolVar
SpCodePresenter>>#browsePoolVarNode:

TODO: remove these methods, deprecation not needed

### LINKED ISSUE #1176: strange: SpCodePresenter>>#doBrowseImplementors reverted to old version
We changed SpCodePresenter>>#doBrowseImplementors in #1150, but the current version in Pharo10 still have the (bad) code
that calls #browseImplementorOfPoolVar

guillep:
Fixes #1176, restores behavior from #1150.

Restore changes in doBrowseImplementors
Remove dead code related to pool variables

From this we realized that browsing implementors on variables is showing their users (wrong!).
Issue: https://github.com/pharo-spec/Spec/issues/1180

### DIFF
--- a/src/Spec2-Code/SpCodePresenter.class.st
+++ b/src/Spec2-Code/SpCodePresenter.class.st
@@ -197,47 +197,6 @@ SpCodePresenter >> bindingOf: aString [
 ]
 
 { #category : #'private - commands' }
-SpCodePresenter >> browseImplementorOfPoolVar [
-
-	"If the selected text is a pool var, list all references and return true. Return false otherwise."
-
-	^ self browsePoolVarNode: [ :astNode | 
-		  self systemNavigation browseHierarchy: astNode binding owningClass ]
-]
-
-{ #category : #'private - commands' }
-SpCodePresenter >> browsePoolVarNode: aBlock [
-
-	"If the selected text is a pool var, list all references  using the block and return true. Return afalse otherwise."
-
-	| astNode selectedText |
-	selectedText := self selectedSelector.
-	selectedText ifNil: [ ^ false ].
-	(interactionModel hasBindingOf: selectedText) ifTrue: [ 
-		astNode := interactionModel ast variableNodes
-			           detect: [ :node | "this way also works without selecting the variable, setting the cursor inside is enogugh"
-				           node start <= self selectionInterval first and: [ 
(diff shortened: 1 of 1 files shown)


MarcusDenker (approved):

src/Spec2-Code/SpCodePresenter.class.st
```diff
@@ -348,14 +307,12 @@ SpCodePresenter >> doBrowseHierarchy [
...
-	| result selectionIsPoolVar |
 
-	selectionIsPoolVar := self browseImplementorOfPoolVar.
-	selectionIsPoolVar ifTrue: [ ^ self ].
-	 
+	| result |
 	result := self
-		browseSelectedSelectorIfGlobal: [ :global | global browse ]
-		ifNotGlobal: [ :selector | self systemNavigation browseAllImplementorsOf: selector ].
+		          browseSelectedSelectorIfGlobal: [ :global | global browse ]
```
MarcusDenker, line 313: this needs to be "global read browse", else we will browse the class of the global variable..

MarcusDenker: (this was already done but lost, see https://github.com/pharo-spec/Spec/pull/1138 )

guillep:
Ok, I'll update the PR.

src/Spec2-Code/SpCodePresenter.class.st
guillep: Well, actually it does not matter for this PR. the ifGlobal block is ignored inside `browseSelectedSelectorIfGlobal: globalBlock ifNotGlobal: nonGlobalBlock`. So it does not matter what we put into that block, it's never executed.

We should solve that as part of #1180.

guillep:
Nope, I'll not :)

That change does not matter for this PR. The ifGlobal block is ignored inside `browseSelectedSelectorIfGlobal: globalBlock ifNotGlobal: nonGlobalBlock`. So it does not matter what we put into that block, it's never executed.

We should solve that as part of #1180.

PR #1181, opened 2021-10-19
Scrollable layout example

StevenCostiou:


Ducasse:
From a naming point of view I would like to have 

SpExecutableLayoutExamplePage
-> SpExecutableLayoutExamplePresenter

because I could not understand what was a page up to the point where I read the class definition.

StevenCostiou:
I changed the naming to `SpScrollableLayoutExample` I have no idea why it was named "executable" instead of "scrollable" (because of autocompletion? but how? anyway...).

I also switched the names between the page and the presenter, because I inverted them.

For the *Page* naming, I followed the convention that seem to be in place for the spec examples.
The page seems to be the object that is retrieved by the demo presenter to show examples, while the presenter is the presenter displayed by the page.

PR #1188, opened 2021-10-22
Removed subclassResponsibility call on defaultSpec

### LINKED ISSUE #1187: Remove the shouldBeImplemented alert for defaultSpec on class side
That alert appears when you don't implement the method `defaultSpec` on the class side of a presenter. As is no longer necessary, because you can initialize your presenters on the instance side, there is no need of the alert. 

Capture d’écran 2021-10-22 à 16 41 44

jordanmontt:
Fixes #1187

### DIFF
--- a/src/Spec2-Core/SpPresenter.class.st
+++ b/src/Spec2-Core/SpPresenter.class.st
@@ -115,6 +115,8 @@ SpPresenter class >> currentApplication [
 SpPresenter class >> defaultSpec [
 
 	"Override to initialize the layout in a non-dynamic way"
+
+	^ nil
 ]
 
 { #category : #accessing }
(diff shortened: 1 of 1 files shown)


src/Spec2-Core/SpPresenter.class.st
```diff
@@ -114,7 +114,7 @@ SpPresenter class >> currentApplication [
 { #category : #specs }
 SpPresenter class >> defaultSpec [
 
-	^ self subclassResponsibility
+	"Override to initialize the layout in a non-dynamic way"
```
estebanlm, line 117: you need to answer nil to make this correct :)

estebanlm (changes requested):

src/Spec2-Core/SpPresenter.class.st
jordanmontt: ok, done it

estebanlm:
@guillep I still have this tests failing. Any hint?

PR #1210, opened 2021-11-29
Use a defaultLayout method to define layouts

demarey:
no more need to explicitly set the layout, you just need to provide a layout in a #defaultLayout method in the presenter. #defaultSpec methods on class side should be avoided.


estebanlm:
errors seems somehow related?

PR #1211, opened 2021-11-30
proposition of API for grid layout

demarey:


estebanlm:
I will close this until the changes we talked about are done, @demarey

PR #1217, opened 2021-12-10
Add a text input field with validation

demarey:


src/Spec2-Core/SpAbstractValidation.class.st
```diff
@@ -0,0 +1,38 @@
+"
+I represent a validation to apply on a presenter.
+I hold the logic to validate the presenter and provide a message if the validation fails.
+If you want to provide a custom validation, subclass me.
+"
+Class {
+	#name : #SpAbstractValidation,
```
astares, line 7: @demarey 
Please implement a class side method #isAbstract to return true.

There is now even a refactoring for that:

image

demarey: Thanks. I was not aware of it.

demarey:
already included in #1240

PR #1220, opened 2022-01-04
Fix for Pressing cmd+m tries to access the senderes instead of the implementations in Pharo 10 #10671

### LINKED ISSUE #10671: Pressing cmd+m tries to access the senderes instead of the implementations in Pharo 10
Capture d’écran 2021-12-20 à 16 00 22

I the screenshot I am pressing `cmd+m` put but is showing me that are no senders.

On Mac, Pharo 10

MarcusDenker:
lookup variable without decaring a new one in case this cose is run with a playground scope as the parent scope

Fixes https://github.com/pharo-project/pharo/issues/10671


MarcusDenker:
The PR needs a lookupVar:declare: method in SystemDictionary, I will do that as it's own PR for Pharo10 (as it makes no sense to have it as an extension for spec).

this PR can only be merged after that is done

PR #1224, opened 2022-01-12
Fix-Deprecated-sends

MarcusDenker:
When reading the build log I found some auto-transforms happening when running the tests

This PR commits these transforms

### DIFF
--- a/src/Spec2-Adapters-Morphic/SpMorphicWorldAdapter.class.st
+++ b/src/Spec2-Adapters-Morphic/SpMorphicWorldAdapter.class.st
@@ -41,5 +41,5 @@ SpMorphicWorldAdapter >> newContainerMorph [
 SpMorphicWorldAdapter >> title: aString [
 	"Update the host window title"
 
-	DisplayScreen hostWindowTitle: aString
+	self currentWorld worldState worldRenderer windowTitle: aString
 ]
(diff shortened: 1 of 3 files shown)


MarcusDenker:
After doing this auto-transform, two transformed tests now fail:

 - #testUnselectedBlockExecutedWhenBecomeUnselected 
 -  #testUnselectedBlockNotExecutedWhenBecomeSelected

estebanlm:
that's weird.

demarey:
The API is not coherent.
A toggle button is now activated or not but the toggle button API uses `#beSelected`,  `#beUnselected` and `#isSelected` as well as a `selected` inst. var.
The change of `#whenSelectedDo` for `#whenActivatedDo`is not valid alone. We should change all or nothing

demarey:
not really

```smalltalk
whenDeactivatedDo: aBlock
  "Inform when the button has been selected. 
   `aBlock` receives zero arguments."

  self whenChangedDo: [ :newSelectedValue | 
    newSelectedValue ifTrue: [ aBlock value ] ]

whenActivatedDo: aBlock
  "Inform when the button has been selected. 
   `aBlock` receives zero arguments."

  self whenChangedDo: [ :newSelectedValue | 
    newSelectedValue ifTrue: [ aBlock value ] ]
```
The implementation of `#whenDeactivatedDo:` and `#whenActivatedDo:`are the same!
I will commit the fix in https://github.com/pharo-spec/Spec/pull/1225

estebanlm:
hah. Good catch. 
The problems of copy & pasting... :)

MarcusDenker:
I close this as the other PR committed them all

PR #1225, opened 2022-01-12
Clean old spec1 API (step 1)

demarey:
All references to Specs will be replaced by layouts.
Also methods named *WithSpec will have the WithSpec suffix removed (not useful)


Ducasse:
I love it!!!!

PR #1228, opened 2022-01-14
fix  'Depracted10' dependencies in baseline

demarey:


demarey:
error not related.
Merge to fix Pharo build

PR #1239, opened 2022-01-24
Fix: Evaluating code in the inspector incorrectly uses binding of on the receiver

### LINKED ISSUE #312: Evaluating code in the inspector incorrectly uses binding of on the receiver
Using binding of (see screenshot) is a potential cause of weird errors.
If we are evaluating code in the context of a dictionary, the dictionary is used for variable lookup, even though its keys may not be names, or we are not intending to.

image

See for example the case below:

image

If I'm on a dictionary, variables are looked up in the dictionary, producing a DNU instead of a more decent compiler error because the name is not defined.

MarcusDenker:
- move bindingOf: method from SpCodeNullInteractionModel to the super class
-  SpCodeObjectInteractionModel can use those

This is tested via SpCodeInteractionModelTest and subclasses

fixes https://github.com/pharo-spec/NewTools/issues/312

### DIFF
--- a/src/Spec2-Code/SpCodeInteractionModel.class.st
+++ b/src/Spec2-Code/SpCodeInteractionModel.class.st
@@ -55,10 +55,10 @@ SpCodeInteractionModel >> doItReceiver [
 	^ nil
 ]
 
-{ #category : #testing }
+{ #category : #binding }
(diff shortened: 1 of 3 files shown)


MarcusDenker:
(name of the PR is wrong as I got confused by the multiple  issue trackers)

PR #1240, opened 2022-01-24
add icetray like presenters

demarey:
add a text input with validation (required field).
add a presenter to present a list of options and a dedicated panel (mac style) with optional validation

### DIFF
--- a/src/Spec2-Adapters-Morphic/SpApplication.extension.st
+++ b/src/Spec2-Adapters-Morphic/SpApplication.extension.st
@@ -1,6 +1,19 @@
 Extension { #name : #SpApplication }
 
 { #category : #'*Spec2-Adapters-Morphic' }
+SpApplication >> addStylesheet: aStylesheet [
+
+	self ensureConfiguration.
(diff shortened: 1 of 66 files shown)


estebanlm (changes requested):
I really do not want any of this as part of a `Spec-Core` package. It is not core! :)

Ducasse:
We should put it in Common-widgets

demarey:
> I really do not want any of this as part of a `Spec-Core` package. It is not core! :)

Why wouldn't it be core? 
All other widgets are there. And these widgets are not really different from others. I would keep them in Core so that people can find them easily, aside other presenters

estebanlm:
because this are not core widget but composed ones. 
no, is not core, please move them to a separated package :)
In fact... there is already a `Spec2-CommonWidgets` package which is seriously more appropriate.

demarey:
> because this are not core widget but composed ones. no, is not core, please move them to a separated package :) In fact... there is already a `Spec2-CommonWidgets` package which is seriously more appropriate.

`Spec2-CommonWidgets` package  is also a good place. Let's do that

PR #1249, opened 2022-02-04
This fixes a non-categrorize method in SpIndeterminatedProgressBarState

MarcusDenker:
all the other implementors are in Spec2-Adapters-Morphic, so I moved this one there, too.

last step (I hope) for https://github.com/pharo-project/pharo/issues/9705


MarcusDenker:
as this is just a recategorization, I merge it

PR #1254, opened 2022-02-23
Fix "Pointers to" in the inspector

MarcusDenker:
needs another change in Newtools


MarcusDenker:
The newtools PR was merged. We need to merge this, too

PR #1257, opened 2022-03-08
Refactoring the Spec in-image examples 

jordanmontt:
Done things

- Created dynamic layout example
- Moved all defaultLayout methods from class side to instance side
- Recategorised the classes inside the package Spec2-Examples
- Reordered the examples
- Added filtering list example
- Moved the form example to the right package
- Refactored the transmission example and added it to the spec demo presenter

Enregistrement-de-lecran-2022-03

### DIFF
--- a/src/Spec2-Examples/SpApplicationWithToolbar.class.st
+++ b/src/Spec2-Examples/SpApplicationWithToolbar.class.st
@@ -38,7 +38,7 @@ SpApplicationWithToolbar >> addItemTo: aGroup [
 SpApplicationWithToolbar >> defaultLayout [
 
 	^ SpBoxLayout newTopToBottom
-		add: menu withConstraints: [ :constraints | constraints height: self class toolbarHeight ];
+		add: menu expand: false;
 		add: text;
 		yourself
 ]
--- a/src/Spec2-Examples/SpApplicationWithToolbar.class.st
+++ b/src/Spec2-Examples/SpApplicationWithToolbar.class.st
@@ -13,14 +13,6 @@ Class {
 	#category : #'Spec2-Examples-Standalone'
 }
 
-{ #category : #layout }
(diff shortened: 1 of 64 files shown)


Ducasse:
Oh this is great!!!

src/Spec2-Examples/SpApplicationWithToolbar.class.st
```diff
@@ -42,6 +34,15 @@ SpApplicationWithToolbar >> addItemTo: aGroup [
 	self build
 ]
 
+{ #category : #layout }
+SpApplicationWithToolbar >> defaultLayout [
+
+	^ SpBoxLayout newTopToBottom
+		add: menu withConstraints: [ :constraints | constraints height: self class toolbarHeight ];
```
estebanlm: in fact... add:height: would be not correct here (`add: ... expand: false` would be)

estebanlm (changes requested):
thanks for this, is very needed!
can you replace the bad usages of add:withConstraints: to fix a height with add:expand: false, as a last step?

src/Spec2-Examples/SpApplicationWithToolbar.class.st
jordanmontt: Ok, I'll do that

jordanmontt:
I push changes in my branch but the PR is not updating... Do you have any idea? @estebanlm

jordanmontt:
@estebanlm I don't know what happened but I made a frivolous commit and now everything is updated. I replaced all `add:withConstraints:` and also deleted the unnecessary accessor methods 🙂

estebanlm:
you left an unused variable at: SpDemoCheckboxesPresenter :)

jordanmontt:
Deleted :)

Do you find something else that can be improved?

PR #1266, opened 2022-03-28
Do not refresh UI too often when using progress presenter.

demarey:
Without this change performance of the work block will be very impacted!

### DIFF
--- a/src/Spec2-Adapters-Morphic/SpMorphicProgressBarAdapter.class.st
+++ b/src/Spec2-Adapters-Morphic/SpMorphicProgressBarAdapter.class.st
@@ -11,7 +11,8 @@ Class {
 		'progressBar',
 		'progressLabel',
 		'lastRefresh',
-		'refreshRateInMs'
+		'refreshRateInMs',
+		'refreshBlock'
 	],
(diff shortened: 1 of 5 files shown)


estebanlm:
I have mixed feelings with the idea of exposing an attribute to control the refresh rate. Why not to make it a minimum in the adapter?

demarey:
> I have mixed feelings with the idea of exposing an attribute to control the refresh rate. Why not to make it a minimum in the adapter?

There is a default reasonable value (250 ms). As we cannot guess all use cases, it is better to let people choose according to their needs: refresh only each 5 seconds (really slow job), or refresh each 100 ms (fast job).
The refresh rate really depends on what you will do.

estebanlm:
yeah, you are answer me with a tautology, precisely what I have mixed feelings is to let the users choose ;)
anyway, tests are failing.

estebanlm:
by the way, the refresh rate has sense if this is a minimum, not a forced rate.

demarey:
I'm looking at the failing test.
It is failing because tests are not ran in the UI thread when using headless mode (they are run in UI thread when in headfull mode)

PR #1290, opened 2022-06-28
Add drop list column support in tables

demarey:


PalumboN (approved):

astares:
Nice!

MarcusDenker:
is the branch correct? This is now dev-1.0, but there is a "Pharo11" branch, too

demarey:
after discussion with @estebanlm , better to be on pharo11. I will change it.

PR #1295, opened 2022-07-19
Add Text Input Field table column

jordanmontt:
This PR adds a text input field column

image


estebanlm:
@jordanmontt this is not needed in general. String table columns are already "editable" or "not editable". 
In case you still want the power of the text input presenter, this extension has to be added under `Spec2-Morphic` package as this is not directly supported on Gtk3 (of course is implementable, but no time to do it right now, and there is not a requirement).

jordanmontt:
Hello, yes you're right I'll close this PR

PR #1298, opened 2022-07-29
[fix] Pharo issue # 11390: Browsing implementors/senders from source code editors doesn't browse expected selector

### LINKED ISSUE #11390: Browsing implementors/senders from source code editors doesn't browse expected selector
**Bug description**
Pharo_bvVjhjRDpy
In the screen recording, I want to browse implementors of `#readParametersFromContext:`, but Pharo instead looks for `#prepareNewCommand:readParametersFromContext:` (which doesn't exist)
It works when explicitly selecting the selector, but it shouldn't be needed

**To Reproduce**
1. Create CmdMenuGroup1, subclass of CmdMenuGroup
2. On class side, define
```
sourceCodeMenuActivation

	<classAnnotation>
	^ SycSourceCodeMenuActivation
		  byItemOf: ClyQueryMenuGroup
		  for: ClySourceCodeContext
```
4. Right click in a source code editor > debug

**Version information:**
 - OS: Windows/Linux
 - Version: Pharo-11.0.0+build.95.sha.554b538cd430662e26bd8ae330e02a5743283451 (64 Bit) (and pharo 10)

hogoww:
Fix https://github.com/pharo-project/pharo/issues/11390
The problem comes from the fact that we are (before this fix) not using information about the string we get, unlike the previous version in RubCodeEditor.
Therefore we do not know when this is a method and when this is not, and this fails for keyword methods.

### DIFF
--- a/src/Spec2-Code-Tests/SpCodePresenterTest.class.st
+++ b/src/Spec2-Code-Tests/SpCodePresenterTest.class.st
@@ -596,7 +596,7 @@ SpCodePresenterTest >> testNilTestSelectedSelectorDoesNotBreakExecution [
 ]
 
 { #category : #'tests - command support' }
-SpCodePresenterTest >> testSelectedSelector [
+SpCodePresenterTest >> testSelectedSelectorForMethodEditing [
 
 	"Testing border cases to avoid breaking execution. The selector extraction logic should be tested in the extractor class"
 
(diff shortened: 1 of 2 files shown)


hogoww:
Bump :D

carolahp:
I have tried this solution and it works, fixing [issue # 11390](https://github.com/pharo-spec/Spec/pull/1298/commits/6d86d74973363ed1ff86c10b08fc9c171f8aebf1)

PR #1300, opened 2022-08-01
Use stateless compiler API

dionisiydk:
Use stateless compiler API. 
The API was always there (evaluate: source vs source: source; evaluate). So that it's fully compatible any compiler version


estebanlm:
what's the gain of this?

dionisiydk:
> what's the gain of this?

See the main PR: https://github.com/pharo-project/pharo/pull/11482 . It's better to discuss it there if you will have more questions.

dionisiydk:
I will close it for now because the diff is confusing now

PR #1306, opened 2022-09-08
latest changes

estebanlm:


estebanlm:
the error is in Pharo11, not Spec

PR #1307, opened 2022-09-08
add removeAllBindings api

estebanlm:


estebanlm:
again, a problem on Pharo11, not here.

PR #1312, opened 2022-09-20
latest changes

estebanlm:


Ducasse:
BTW pablo said that he is working on an Icon class to support different sizes and more.

PR #1314, opened 2022-09-21
fix testIconProvider

estebanlm:


Ducasse:
Esteban this is strange because I see this 
Capture 2022-09-21 at 21 20 27

estebanlm:
yeah, I removed that line because it does not has sense anymore.
Now "command iconProvider" should be "command context application". I could add that validation but I just looked for a quick fix :P

Ducasse:
But then why not removing the test. I do not see the point to have a test without assertion.

estebanlm:
there are asserts :)
the complete test is like this: 

```Smalltalk
testIconProvider
	
	command iconProvider: self.
	
	self assert: command iconProvider equals: self.
	
	command iconName: #banana.
	self assert: (command icon) equals: self banana
```

PR #1317, opened 2022-10-05
fix problems after removal of spec1 deprecation (+others)

estebanlm:


estebanlm:
this will be fixed on merge (since is in newtools, the problem, and I just fixed it)

PR #1324, opened 2022-10-28
[improvement] ask the object what to select when adding a presenter to a MillerCollumnPresenter

hogoww:
One common usage is to remove association inspecting in the inspector.
I'm not a fan of the name of the selector, nor the package in which it is added.
But I trust @estebanlm to correct me !

Dictionary inspection with change:
Screenshot 2022-10-28 at 15 28 12

### DIFF
--- a/src/Spec2-Core/Association.extension.st
+++ b/src/Spec2-Core/Association.extension.st
@@ -1,0 +1,6 @@
+Extension { #name : #Association }
+
+{ #category : #'*Spec2-Core' }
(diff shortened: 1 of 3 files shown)


hogoww:
I see that Guille fixed it already, in another manner.
https://github.com/pharo-spec/NewTools/pull/279
Feel free to discard this PR then, unless you think it's worth having such a mechanism

jecisc:
Thanks @hogoww, since this is pretty specific I’ll discard it for now. :)

PR #1328, opened 2022-11-21
add missing method

estebanlm:


estebanlm:
this fixes https://github.com/pharo-spec/NewTools/issues/424

Issue #1341, opened 2022-12-22
Add source and target labels to SpChooserPresenter


n1toxyz:
I would like to propose to add two labels to SpChooserPresenter, which take the role of headings above the source and target lists. 

These labels give more context to the widget. As it stands now it isn't clear which side of the widget contains the chosen items just by looking at it. The only way to tell for the user is to hover over one of the buttons and look at the popup text to see if the button adds or removes the item, at least as far as I can tell.

Of course the labels can be made optional and a method to disable them may be provided inside the class, but I think they should be enabled by default.

A similar layout is already used in the Finder's package chooser dialog with DialogItemsChooserUI.

### Screenshots:

**Left:** Old SpChooserPresenter (without labels); **Right:** New SpChooserPresenter (with labels)

![chooser_labels](https://user-images.githubusercontent.com/29785597/209230630-a5447369-abba-452e-9f47-3e424d088e69.png)

**Package chooser dialog of the Finder tool**

![package_chooser](https://user-images.githubusercontent.com/29785597/209231922-9ae25d76-88ef-458c-895c-901d37d480ed.png)

Without the labels it would not be obvious, which side contains the selected packages.

demarey:
It would be nice but I would use "Available items" and "Selected items".

Ducasse:
Christophe can you integrate? I cannot

demarey:
Thank you

PR #1342, opened 2022-12-23
Add source and target labels to SpChooserPresenter ( #1341)

n1toxyz:
Adds source and target labels to SpChooserPresenter as well as tests for them. Also sets the protocol of defaultLayout to layout as recommended.


Ducasse:
Thanks! This is a good addition.

demarey:
error not related to the PR

Issue #1275, opened 2022-05-10
SpTextPresenter whenScrollValueChangedDo: not reacting to scrollbar


kasperosterbye:
It seems like the scrollbar of SpTextPresenter (morphic version) is not connected to whenScrollValueChangedDo:.
I could not get it to work, and the example SpScrollSyncExample is not working either (the method syncMorphPosition is not called).

hernanmd:
I fixed this in Pharo 11 by specializing scrollerOffsetChanged and registerScrollChanges in SpRubScrolledTextMorph. However I don't know if it's correct enough to submit as PR @estebanlm ? :

Here are the methods for **SpRubScrolledTextMorph**:

```smalltalk
scrollerOffsetChanged

	super scrollerOffsetChanged.
	self announcer
		announce:
			(PaneScrolling new
				step: self scrollPane hScrollbarValue @ self scrollPane vScrollbarValue;
				scrollPane: self scrollPane;
				yourself)
```

```smalltalk
registerScrollChanges: aSelector

	self announcer 
		when: PaneScrolling 
		send: aSelector
		to: self model
```

And **SpScrollSyncExample** also needs to be fixed:

```smalltalk
syncMorphPosition

	currentVerticalScrollPosition = text scrollValue y
		ifTrue: [ ^ self ].
	currentVerticalScrollPosition := text scrollValue y.
	self needRebuild: false.
	text needRebuild: false.
```

```smalltalk
initializePresenters
	text := self newText.
	sideRuler := self newMorph.

	sideRuler
		morph:
			(Morph new
				color: Color yellow;
				extent: 50 @ 500;
				useGradientFill;
				hResizing: #spaceFill;
				vResizing: #spaceFill;
				yourself).

	text whenScrollValueChangedDo: [ :newPosition | 
		self traceCr: 'scroll position: ' , newPosition asString. 
		self syncMorphPosition ].
	text text: self loremIpsum.
	text registerEvents.
```

hernanmd:
This issue can be closed now

PR #1344, opened 2023-01-05
Issue 1275: Notify scroll changes

hernanmd:
Specialized scrollerOffsetChanged and registerScrollChanges: to notify changes when scrolling in Morphic (Issue 1275 https://github.com/pharo-spec/Spec/issues/1275 SpTextPresenter whenScrollValueChangedDo: not reacting to scrollbar).

Updated SpScrollSyncExample methods to display the scroll position when the scroll value changes.

### DIFF
--- a/src/Spec2-Morphic-Examples/SpScrollSyncExample.class.st
+++ b/src/Spec2-Morphic-Examples/SpScrollSyncExample.class.st
@@ -72,8 +72,11 @@ SpScrollSyncExample >> initializePresenters [
 				vResizing: #spaceFill;
 				yourself).
 
-	text whenScrollValueChangedDo: [ :newPosition | self syncMorphPosition ].
-	text text: self loremIpsum
+	text whenScrollValueChangedDo: [ :newPosition | 
+		self traceCr: 'scroll position: ' , newPosition asString. 
+		self syncMorphPosition ].
+	text text: self loremIpsum.
+	text registerEvents.
 ]
 
 { #category : #private }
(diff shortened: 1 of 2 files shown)


src/Spec2-Morphic-Examples/SpScrollSyncExample.class.st
```diff
@@ -88,12 +91,13 @@ SpScrollSyncExample >> sideRuler [
 
 { #category : #private }
 SpScrollSyncExample >> syncMorphPosition [
+
 	currentVerticalScrollPosition = text scrollValue y
 		ifTrue: [ ^ self ].
 	currentVerticalScrollPosition := text scrollValue y.
 	self needRebuild: false.
```
estebanlm: `needRebuild:` call should not be needed anymore in both cases

src/Spec2-Morphic-Examples/SpScrollSyncExample.class.st
```diff
@@ -72,8 +72,11 @@ SpScrollSyncExample >> initializePresenters [
 				vResizing: #spaceFill;
 				yourself).
 
-	text whenScrollValueChangedDo: [ :newPosition | self syncMorphPosition ].
-	text text: self loremIpsum
+	text whenScrollValueChangedDo: [ :newPosition | 
+		self traceCr: 'scroll position: ' , newPosition asString. 
+		self syncMorphPosition ].
+	text text: self loremIpsum.
+	text registerEvents.
```
estebanlm: `registerEvents` should not be called, is part of the initialization process.
(and it will introduce leaks if you call it)

estebanlm (changes requested):
those details need to be fixed :)

src/Spec2-Morphic-Examples/SpScrollSyncExample.class.st
hernanmd: Ok done. Let me know any further changes.

estebanlm (approved):

src/Spec2-Morphic-Examples/SpScrollSyncExample.class.st
hernanmd: I saw that there is a problem with the [CI](https://github.com/pharo-spec/Spec/actions/runs/3846092231/jobs/6551116716) but I don't quite understand what it is:

```
Could not register the tested repository with Iceberg.

NotFound: object not found - no match for id (8935da6bbbb6fb3767cd26099b6acd5f311ae911)
```

PR #1360, opened 2023-03-04
Adding textChanged notification to SpStringTableColumn

colorow:
This is a fix for issue #12884 on Pharo-project-pharo

Currently an SpStringTableColumn can be set to editable, but the changes are only sent to the table #items when the user does a carriage return.

This change adds a text changes event which will sent updates to the table at each keystroke.

I could not figure out a way to write a test case for this, as the editing occurs in a popup text input over the string column.  But I did include a change to SpTablePresenter >> exampleWriteableTable which excercises the change.  To run this code:
1. Open a Transcript.
2.  Execute "SpTablePresenter exampleWriteableTable" in a Playground.  
3. Edit and cell in the left hand column.  Changes should show in the Transcript.


jordanmontt:
Hello, what is needed to accept this PR? @estebanlm

estebanlm:
@jordanmontt  first of all this should now be going to Pharo12 branch, not to Pharo11 ;)

jordanmontt:
It seems that you changed the target branch :)

estebanlm:
lol, this thing never went to the trunk :P 
checking now if we can add it.

PR #1363, opened 2023-03-23
SpCodeInteractionModel>>#notify: Avoid crashing

privat:
if the message is exactly `'->'`, it will try to allocate a -1 sized string, thus crash.

I copy-pasted these two lines without checking the arithmetic, and it bit me. cf https://github.com/pharo-project/pharo/issues/13107

### DIFF
--- a/src/Spec2-Code/SpCodeInteractionModel.class.st
+++ b/src/Spec2-Code/SpCodeInteractionModel.class.st
@@ -66,23 +66,23 @@ SpCodeInteractionModel >> isScripting [
 SpCodeInteractionModel >> notify: message at: location in: code [
 	| currentSelection stripMessage |
 
-	"self owner withAdapterDo: [ :anAdapter | 
+	"self owner withAdapterDo: [ :anAdapter |
 		anAdapter notify: message at: location in: code ]"
 
-	stripMessage := (message endsWith: '->')
+	stripMessage := (message endsWith: ' ->')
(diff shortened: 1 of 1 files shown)


src/Spec2-Code/SpCodeInteractionModel.class.st
```diff
@@ -66,23 +66,23 @@ SpCodeInteractionModel >> isScripting [
 SpCodeInteractionModel >> notify: message at: location in: code [
 	| currentSelection stripMessage |
 
-	"self owner withAdapterDo: [ :anAdapter | 
+	"self owner withAdapterDo: [ :anAdapter |
 		anAdapter notify: message at: location in: code ]"
 
-	stripMessage := (message endsWith: '->')
+	stripMessage := (message endsWith: ' ->')
 		ifTrue: [ message allButLast: 3 ]
```
demarey, line 73: I wonder if the right change is not `message allButLast: 2` but I do not know what kind of messages this method expect

estebanlm: .. or `(allButLast: 2) trimRight` ?

privat: There are only two real senders, one ending its messages with `' ->'` and the without them, so it does not change that much. It's not that important, but the bad math feels just bad (and did hurt me when I used only '->' as message by mistake).

MarcusDenker (approved):

PR #1370, opened 2023-03-24
Fix scrollTo: on SpScrollableLayout not working

hernanmd:
Add SpMorphicScrollableAdapter>>scrollTo:
Add example in class documentation.


estebanlm:
Sorry for coming to this so late.
there is a problem with this: in the Gtk implementation (which arrived first, this time :P) the scroll value is not a number between 0 and 1 but a real offset. 
To keep the same criteria, you need to receive also a real offsent then transform it into the percent value (which is what the 0..1 value is) in the adapter.

estebanlm (changes requested):
what I set in comment ;)

hernanmd:
@estebanlm 

I've seen the changes should be applied to `GtkScrollableAdapter>>scrollTo:`.
I should get then the maximum offset of the adapter which holds the scroller?
Any hints?

hernanmd:
> Sorry for coming to this so late. there is a problem with this: in the Gtk implementation (which arrived first, this time :P) the scroll value is not a number between 0 and 1 but a real offset. To keep the same criteria, you need to receive also a real offsent then transform it into the percent value (which is what the 0..1 value is) in the adapter.

I don't know how to adapt this method for the GTK version (I tried and it is not working).
Could we have this integrated anyway as this is needed for the NewTools File Browser? Since @jecisc worked in the File Browser too what do you think ?

Ducasse:
Hernan is it not that the adpater should do the conversion following the way it is done in GTK backend. 
We specify X in the presenter and x goes to GTKbackend and a transformation of (X) to the morphic one?

estebanlm:
do not care about Gtk backend. 
I was thinking the opposite: you need to do the morphic backend to accept a real offset and internally (in the adapter) transform it into a percentage (since you will have the data needed there).

Ducasse:
Ok thanks. Now this is clearer.

Ducasse:
So what do we do with this one.

estebanlm:
I asked for some changes...

hernanmd:
> I asked for some changes...

Sorry, I never got quite right how to implement them.
Do you think we could sync this week?

Ducasse:
@hernanmd | @estebanlm 
can you take action? @estebanlm if you know and it takes you 10 min can you do it because
I have the impression that @hernanmd does not understand or explain to him.

estebanlm:
yeah, we will see it as soon as we are both in the same place ;)

Ducasse:
ping! @hernanmd @estebanlm

Ducasse:
Broken test is unrelated 

```
 ✗ #testUndeclared (165ms)
TestFailure: Found undeclared Variables: 
StSystemReporter in: SmalltalkCIPharo9>>#imageInfo
Gofer in: SCIGoferLoadSpec>>#basicLoadProjectOn:
ReleaseTest(TestAsserter)>>assert:description:resumable:
ReleaseTest(TestAsserter)>>assert:description:
ReleaseTest>>testUndeclared ...assert: remaining isEmpty description: description
ReleaseTest(TestCase)>>performTest
```

so I merge

jecisc:
We have 2 failing tests about scrolling in the latest Pharo build. I think they come from here

PR #1374, opened 2023-04-04
[WIP] Change announcement api

PalumboN:
We changed the Announcement API

```
when: event do: action
```
by
```
when: event do: action for: subscriber
```

See https://github.com/pharo-project/pharo/pull/13270 for more details

### DIFF
--- a/src/Spec2-Adapters-Morphic/SpAbstractMorphicAdapter.class.st
+++ b/src/Spec2-Adapters-Morphic/SpAbstractMorphicAdapter.class.st
@@ -684,11 +684,10 @@ SpAbstractMorphicAdapter >> vSpaceFill [
 ]
 
 { #category : #protocol }
-SpAbstractMorphicAdapter >> when: anAnnouncement do: aBlock [
+SpAbstractMorphicAdapter >> when: anAnnouncement do: aBlock for: aSubscriber [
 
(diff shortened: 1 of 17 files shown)


src/Spec2-Adapters-Morphic/SpAbstractMorphicAdapter.class.st
```diff
@@ -683,10 +684,10 @@ SpAbstractMorphicAdapter >> vSpaceFill [
 ]
 
 { #category : #protocol }
-SpAbstractMorphicAdapter >> when: anAnnouncement do: aBlock [
```
jecisc, line 686: Maybe we should do as in the PR of Pharo and keep the old version deprecated

jecisc (changes requested):

jordanmontt:
The code seems ok :) there are just some comments to be fixed

PalumboN:
I close this PR because:
- [The change](https://github.com/pharo-project/pharo/pull/13590) in the Pharo repo was merged to pharo-12
- I wrote a deprecation rule for that, so the code should be ~~already changed in pharo-12 branch~~ [in this PR](https://github.com/pharo-spec/Spec/pull/1401) (in this repo)
- Probably some improvements could be made on this side too, but I prefer to do it in another PR :)

PR #1379, opened 2023-04-20
SpCodeInteractionModel: add compile and test it

privat:
https://github.com/pharo-project/pharo/pull/13498 did solve a regression with dolt evaluation in the debugger.

This PR move a `compiler` helper to the code interaction model (used by SpCodePresenter) and add (and fix) tests to prevent future issues.

Because of the regression, I assume that #13498 should be merged first.


privat:
Tests are OK (I'm not responsible for class comments as I did not add a class)

MarcusDenker (approved):

PR #1380, opened 2023-04-25
Half of a fix for "right-clicking empty Iceberg pane leads to message…

Mr-Dispatch:
… being sent to nil".

The other half is in Iceberg.


Mr-Dispatch:
The other half is at: https://github.com/pharo-vcs/iceberg/pull/1698

PR #1393, opened 2023-05-16
Move-isObservableSlot

MarcusDenker:
#isObservableSlot was part of a deprecated package, move it to the correct package


jecisc:
My guess is that the problem is this:
- The Pharo image get bootstraped and loads the VariableLibrary package with the Pharo vesion of #isObservableSlot 
- Later in the bootstrap the Pharo12 branch of spec is loaded and install its own version of #isObservableSlot 
- The CI here uses this image and unload Spec to reload the new version, its version of #isObservableSlot is removed but the Pharo original version is not installed back. 

So, merging this should be fine.

MarcusDenker:
ah, let's try

PR #1397, opened 2023-05-20
Add class variable to control scaling in SpFontStyle

Rinzwind:
When setting ‘Standard fonts’ in the Settings Browser to the predefined style ‘Very large’, some text becomes too large:

<p align="center">

</p>

This is because besides setting the fonts to a larger point size, the style also sets a larger display scale factor, which SpFontStyle then uses to further scale up the point size. This pull request adds a class variable that allows turning that off:

```smalltalk
SpFontStyle applyDisplayScaleFactor: false
```

<p align="center">

</p>

For additional context, see this pull request: https://github.com/pharo-project/pharo/pull/13767.

### DIFF
--- a/src/Spec2-Adapters-Morphic/SpFontStyle.class.st
+++ b/src/Spec2-Adapters-Morphic/SpFontStyle.class.st
@@ -155,7 +155,7 @@ SpFontStyle >> calculateFontSize [
 	
 	self sizeVariable ifNotNil: [ :aVariable |
 		aVariable value ifNotNil: [ 
-			^ self sizeVariable scaledValue ] ].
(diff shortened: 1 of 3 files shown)


Rinzwind:
Note: I was doubting between [‘nonscaled’](https://en.wiktionary.org/w/index.php?title=nonscaled&oldid=51770554) and [‘unscaled’](https://en.wiktionary.org/w/index.php?title=unscaled&oldid=50833821). Wiktionary says:

> **nonscaled:** Not scaled (resized according to a scale or measure)
>
> **unscaled:** That has not been scaled (climbed)

jecisc (approved):

PR #1398, opened 2023-05-22
cleanup SpCodePresenter>>#findClassFromSelection

MarcusDenker:
SpCodePresenter had methods with "self flag:" that are not used

- findClassFromSelection
- findClassFrom:

This PR just removes that code and the test.

This is nice as this removes one direct use of RBParser


jordanmontt (approved):
Nice cleanup

jecisc (approved):

PR #1399, opened 2023-05-26
[Pharo11] Error Fueling out SpStyleEnvironmentColorProxyTest instances

tesonep:
- As the proxy extends from ProtoObject we have to correctly implement #fuelAccept:, #instVarAt: and #instVarAt:put:

- Add a Test
- Fix https://github.com/pharo-project/pharo/issues/13788


jecisc (approved):

jecisc:
This seems good. 
But if I merge this, it will not trigger a new Pharo 11 generation, so I'm not sure how to do that?
Do we merge and than can you trigger manually a new generation?

estebanlm:
> This seems good. But if I merge this, it will not trigger a new Pharo 11 generation, so I'm not sure how to do that? Do we merge and than can you trigger manually a new generation?

it will not, it requires a new tag version and a change in P11.

also @tesonep, I did a question in P12 PR... what was the answer?

PR #1401, opened 2023-05-26
Updated code following deprecation

### LINKED ISSUE #13807: DeprecationPerformedNotification during build
Running tests of NewTools-Inspector-Tests
DeprecationPerformedNotification: Automatic deprecation code rewrite: The method Announcer>>#when:do: called from SpExecutableLayout>>#whenChildrenAddedDo: has been deprecated. Since there are some block closures (Clean and Constant) without a receiver, the API of announcements was changed to send the subscriber explicitly. 
			We are deprecating this method because it was asking for the receiver of the block to use it as the subscriber.
DeprecationPerformedNotification: Automatic deprecation code rewrite: The method Announcer>>#when:do: called from SpExecutableLayout>>#whenChildrenRemovedDo: has been deprecated. Since there are some block closures (Clean and Constant) without a receiver, the API of announcements was changed to send the subscriber explicitly. 
			We are deprecating this method because it was asking for the receiver of the block to use it as the subscriber.
Running tests of BaselineOfReflectivity
Running t

DurieuxPol:
Fixes [#13807](https://github.com/pharo-project/pharo/issues/13807)
Changes from deprecation


src/Spec2-Layout/SpExecutableLayout.class.st
```diff
@@ -343,6 +341,7 @@ SpExecutableLayout >> whenChildrenRemovedDo: aBlock [
 	self announcer
 		when: SpChildrenRemoved
 		do: aBlock
+		for: aBlock receiver
```
MarcusDenker, line 348: "aBlock receiver" is not good here...  for clean blocks it will be nil

estebanlm: oops I resolved by mistake :P

Ducasse:
I fixed the conflict: it was just formatting. 
And merged.

PR #1402, opened 2023-05-26
fix: exception when opening Spec demo "Object inspector" notebook

ValentinBourcier:
See: https://github.com/pharo-project/pharo/issues/13685

### DIFF
--- a/src/Spec2-Core/SpNotebookPage.class.st
+++ b/src/Spec2-Core/SpNotebookPage.class.st
@@ -13,7 +13,6 @@ Class {
 		'#owner',
 		'#title => ObservableSlot',
 		'#icon => ObservableSlot',
-		'#layoutSpec => ObservableSlot',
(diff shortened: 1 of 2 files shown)


src/Spec2-Core/SpNotebookPage.class.st
```diff
@@ -70,17 +69,6 @@ SpNotebookPage class >> title: aString icon: anIcon provider: aBlock [
...
-SpNotebookPage class >> title: aString icon: anIcon provider: aBlock layoutSpec: aSymbol [
-
-	^ self new 
-		title: aString;
-		icon: anIcon;
-		presenterProvider: aBlock;
-		layoutSpec: aSymbol;
-		yourself
-]
-
```
jecisc, line 83: This method is called in SpDemoNotebookPresenter>>objectInspectorPage I don't know why it was removed

PR #1403, opened 2023-05-30
fix #canAddBindingOf: to not allow synthetic names

MarcusDenker:
fix #canAddBindingOf: to not allow synthetic names (starting with a number) as used in optimized constructs


MarcusDenker:
No, it is better to do this in OCRequestorScope>>#canAddBindingOf:

PR #1406, opened 2023-06-02
Fix Smoke test failing in Pharo

jecisc:


demarey (approved):
the code formatting was more readable before the change

PR #1407, opened 2023-06-06
Update usage of protocols

jecisc:
CompiledMethod>>protocol currently returns a symbol. In the future it will return a real protocol. To help with the change a method #protocolName was introduced. Here I am updating Spec usage of protocols to use this new methods when needed and keep #protocol when it makes sense


demarey:
The change is ok but the code is a  big mess.
There is a lot of code duplication. For example, SpTreePresenter and SpTreeTablePresenter define the same method that could be factorized in SpAbstractTreePresenter.

demarey (approved):

PR #1411, opened 2023-06-12
Simplify #doBrowseImplementors and #doBrowseSenders

MarcusDenker:
#doBrowseImplementors and #doBrowseSenders where a bit hard to follow with the odd #browseSelectedSelectorIfGlobal:ifNotGlobal: method.

This PR intead inlines the logic, which makes it much clearer what happenes


estebanlm (approved):
not sure how this work as a simplification, but ok :)

estebanlm:
but there are a lot of tests failng

MarcusDenker:
For failing tests, I added https://github.com/pharo-spec/Spec/issues/1415
(they are not a result of this PR)

PR #1413, opened 2023-06-13
fix and move SHRBTextStyler

MarcusDenker:
- remove all extension methods of SHRBTextStyler (they are in Pharo after https://github.com/pharo-project/pharo/pull/14011)

- fix one method that was using the deprecated method

this needs to be merged *after* the Pharo level change


MarcusDenker:
as the other pr was merged, I merge this

PR #1427, opened 2023-07-22
SpMorphicBackendForTest>>#doubleClickFirstRowAndColumn:

ironirc:
now uses MouseButtonEvent (instead of MouseEvent)
Essentially doesn't change much, but allows #asPseudoDoubleClickEvent to be sent to the event.
This supports changes made to improve double click handling (e.g. on FTTableMorph)
See https://github.com/pharo-project/pharo/issues/14168


jecisc (approved):

ironirc:
Tests are failing, but I can't figure out why? Can I do something?

Ducasse:
The problem is that I do not know how Spec changes get available in Pharo12 dev.

PR #1428, opened 2023-07-22
[Backport] SpMorphicBackendForTest>>#doubleClickFirstRowAndColumn:

ironirc:
now uses MouseButtonEvent (instead of MouseEvent)
Essentially doesn't change much, but allows #asPseudoDoubleClickEvent to be sent to the event.
This supports changes made to improve double click handling (e.g. on FTTableMorph)
See https://github.com/pharo-project/pharo/issues/14168


ironirc:
hm,  seems I really messed up here, thinking that the 1 method I changed would be picked up as diff, and be applied to Pharo11.
Apparently that's not how it works.
I'll try to do a new PR.

PR #1445, opened 2023-09-03
Reduce UIManager (hidden) dependencies

Ducasse:
Reduce UIManager dependencies.
Other PRs depend on it.


Ducasse:
There is one red tests in integration that is "references" and is taking too much time. 
So not related.

PR #1448, opened 2023-09-08
Skip testDoBrowseMethodReferences until it passes

jecisc:
testDoBrowseMethodReferences is failing because the UI thread is blocked during its execution. I propose to skip it until we have a better solution (like non blocking informs)

I also updated a deprecated method


Ducasse:
Thanks I was working on a notify alternative but it is not fully working.

PR #1452, opened 2023-09-12
Added new subclass with unified view based on SpDiffPresenter

CarolinaMelendez:
SpDiffPresenter displays a **split view**. So I added a sub-class that displays a **unified view** called SpDiffUnifiedPresenter

image

### DIFF
--- a/src/Spec2-Adapters-Morphic/SpMorphicDiffUnifiedAdapter.class.st
+++ b/src/Spec2-Adapters-Morphic/SpMorphicDiffUnifiedAdapter.class.st
@@ -1,0 +1,22 @@
+Class {
+	#name : #SpMorphicDiffUnifiedAdapter,
+	#superclass : #SpMorphicDiffAdapter,
+	#category : #'Spec2-Adapters-Morphic-Base'
+}
+
+{ #category : #factory }
+SpMorphicDiffUnifiedAdapter >> buildWidget [
+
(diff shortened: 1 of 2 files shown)


estebanlm:
but this will not work.
You need to include `StUnifiedDiffChangesMorph` in the correct place.
(and call it `SpUnifiedDiffChangesMorph`) :P

Ducasse:
BTW carolina talk to us if you want to do your bachelor with us at Lille. You can talk to seb.

CarolinaMelendez:
Hi @estebanlm thanks for the review. I see I need to learn more. By the correct place of StUnifiedDiffChangesMorph you mean it should be in the same package together with DiffMorph ?

I would like to add that StUnifiedDiffChangesMorph is not a class I created, I just reuse that class because it already has the unified view there. I don't think I have the right to move the package. There is not a another way that make it works or is necessary to move the the location of StUnifiedDiffChangesMorph ?

jordanmontt:
Hello, what is needed to accept this PR?

estebanlm:
heh... nothing :)
EDIT: I just saw there is a problem with dependencies.

Ducasse:
IAfter 10 min I could not find a way to rerun this PR :(
What a dead interface

Ducasse:
So we will see if it breaks and in this case we will revert the PR

PR #1454, opened 2023-09-19
use-optimzied-ifNil

MarcusDenker:
Two simple changes to make sure that a #ifNotNil:#ifNil:  send is inlined


MarcusDenker:
Changes are in 

SpMorphicTableDataSource>>#elements
SpDatePresenter>>#initialize

The SpMorphicTableDataSource method seems to be called a lot from the inspector (e.g. in the debugger), so avoiding some sends and block creations is nice

demarey:
Why symbols are now serialized as Strings in Tonel?

MarcusDenker:
There is now a conflict, I will redo the PR

PR #1463, opened 2023-10-16
Fix MNU when using component list as input port in custom presenter

hernanmd:
This PR handles a case with transmissions along a component list in custom presenters. There is an issue in transmitTo:transform: method when connecting a transmission between presenters where the target presenter has defined an input port as a `SpListItemPort` with a component list . 

```smalltalk
defaultInputPort 

	^ SpListItemsPort newPresenter: componentListPresenter
```

The PR includes a case with a test to reproduce the problem:

```smalltalk
SpComplexComponentListExample open.
```

To fix it, the PR specializes `transmitTo:transform:` to handle problem with connecting presenters with component list.

### DIFF
--- a/src/Spec2-Core/SpComponentListPresenter.class.st
+++ b/src/Spec2-Core/SpComponentListPresenter.class.st
@@ -83,6 +83,14 @@ SpComponentListPresenter >> presenters: aSequenceableCollection [
 	^ self items: aSequenceableCollection
 ]
 
+{ #category : 'transmitting' }
+SpComponentListPresenter >> transmitTo: aPresenter transform: aValuable [
+	
+	^ self defaultOutputPort
+		transmitTo: aPresenter
+		transform: aValuable
+]
+
 { #category : 'api - events' }
 SpComponentListPresenter >> whenPresentersChangedDo: aBlock [
 	"Inform when the presenter list changed (See `SpComponentListPresenter>>#presenters:`.
--- a/src/Spec2-Tests/SpComplexComponentListExample.class.st
+++ b/src/Spec2-Tests/SpComplexComponentListExample.class.st
@@ -1,0 +1,46 @@
+Class {
+	#name : 'SpComplexComponentListExample',
+	#superclass : 'SpPresenter',
+	#instVars : [
+		'leftPresenter',
+		'rightPresenter',
+		'toolbarPresenter'
+	],
+	#category : 'Spec2-Tests-Utils',
+	#package : 'Spec2-Tests',
+	#tag : 'Utils'
+}
+
+{ #category : 'instance creation' }
+SpComplexComponentListExample class >> open [
+	<example>
+	^ self new open
+]
+
+{ #category : 'initialization' }
+SpComplexComponentListExample >> connectPresenters [ 
+
+	leftPresenter
+		transmitTo: rightPresenter 
+		transform: [ : item | { item label asNumber . (item label asNumber * 2) }  ]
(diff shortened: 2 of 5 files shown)


jecisc (approved):

estebanlm:
sorry but this fix is wrong.

estebanlm:
transmissions do not work between presenters but between ports `OutputPortA` connects to `InputPortB`
the method `transmitTo:` (and family) present in `SpPresenter` are convenience methods to connect the *default output port* of a presenter with the *default input port* of another presenter, in that sense, doing: 
```smalltalk
aPresenter transmitTo: otherPresenter
```
is the same as doing:
```smalltalk
aPresenter defaultOutputPort transmitTo: otherPresenter defaultInputPort
```
but you cannot take a port and connect it to a presenter.

EDIT: basically, you cannot use a presenter as input port.

estebanlm:
I made a revert: 

https://github.com/pharo-spec/Spec/pull/1464

hernanmd:
> transmissions do not work between presenters but between ports `OutputPortA` connects to `InputPortB` the method `transmitTo:` (and family) present in `SpPresenter` are convenience methods to connect the _default output port_ of a presenter with the _default input port_ of another presenter, in that sense, doing:
> 
> ```smalltalk
> aPresenter transmitTo: otherPresenter
> ```
> 
> is the same as doing:
> 
> ```smalltalk
> aPresenter defaultOutputPort transmitTo: otherPresenter defaultInputPort
> ```
> 
> but you cannot take a port and connect it to a presenter.
> 
> EDIT: basically, you cannot use a presenter as input port.

The [documentation](https://github.com/pharo-spec/Spec/blob/Pharo12/spec2.md) isn't exactly clear about what's the difference between "using as" and "connecting".

The documentation mentions that "a transmission connects a presenter's output port with a presenter's input port," whereas you wrote, "transmissions do not work between presenters but between ports." This is contradictory because the documentation implies that the transmission connects presenters indirectly through their ports, whereas you suggest that the transmission works only between ports, not presenters.

Additionally you've mentioned "you cannot take a port and connect it to a presenter," which contradicts the information in the documentation, where it is shown that a port can be connected to a presenter's default input port, as in the example "list defaultOutputPort transmitTo: detail defaultInputPort".

I provided an `connectPresenters` use case where following the documentation does not work. 
What do you suggest to modify in order to run it without errors?

estebanlm:
this part or your case: 
```smalltalk
defaultOutputPort 

    ^ self categoryListPresenter
```
is wrong. You cannot answer a presenter as a port.

demarey:
There is a draft chapter on transmissions *Using transmissions and ports* in the Spec2 book: https://github.com/SquareBracketAssociates/BuildingApplicationWithSpec2/releases/download/latest/Spec2-wip.pdf 

You can simply define `#defaultOutputPort` as follows:
```smalltalk
defaultOutputPort 

    ^ self categoryListPresenter defaultOutputPort
```

hernanmd:
> There is a draft chapter on transmissions _Using transmissions and ports_ in the Spec2 book: https://github.com/SquareBracketAssociates/BuildingApplicationWithSpec2/releases/download/latest/Spec2-wip.pdf
> 
> You can simply define `#defaultOutputPort` as follows:
> 
> ```smalltalk
> defaultOutputPort 
> 
>     ^ self categoryListPresenter defaultOutputPort
> ```

Thanks @demarey ! That's exactly what I was looking for.

PR #1472, opened 2023-11-19
Introduce newProgress in SpApplication. It was missing

Ducasse:
Then since all the newInform and friends were defined as extensions move their users as extensions too (else it does not make sense).


Ducasse:
Once integrated I will fix all the IceTip references to UIManager since I'm the garbage guy.

Ducasse:
I will do it in another way

PR #1478, opened 2023-11-24
1477-Doing-an-informduring-in-SpApplication-call-twice-the-during-block

### LINKED ISSUE #1477: Doing an inform:during: in SpApplication call twice the during block
Doing an inform:during: in SpApplication call twice the during block

tesonep:
Fixing the SpApplication>>#informUser:during: was calling twice openModal and running the during block twice.
Fix #1477


estebanlm (approved):

Ducasse:
Sorry for the bug and tx for the fix

PR #1481, opened 2023-11-30
Commit deprecations transforms

### LINKED ISSUE #1460: SpAbstractPresenter>>#whenBuiltDo: is calling a deprecated method
DeprecationPerformedNotification: Automatic deprecation code rewrite: The method Announcer>>#when:do: called from SpAbstractPresenter>>#whenBuiltDo: has been deprecated

### LINKED ISSUE #1468: SpJob>>#whenStartDo: is calling a deprecated method
The method Announcer>>#when:do: called from SpJob>>#whenStartDo: has been deprecated.

MarcusDenker:
fixes  #1460
fixes #1468


MarcusDenker:
We need to check later as for clean blocks, the for: will be nil. We should check if that leads to any problems, maybe clients need to use a versions of these methods with an explicit for: argument.

PR #1488, opened 2023-12-09
Add a Spec component list example

hernanmd:
This PR add a presenter example in Pharo 12 to show the basic usage of a component list. The component list holds items which are collected as presenters, each time an item is clicked a detail list is updated.

Add also tests for opening and selection.


Ducasse:
Thanks!
In addition we should probably add a test that checks that all the examples are actually at least opening.

Ducasse:
No idea why the integration is failing.

```
Downloading the latest 120 Image:
    http://files.pharo.org/get-files/120/pharoImage-x86_64.zip

Error with status code 56:
286 pharo::prepare_image /home/runner/.smalltalkCI/pharo/run.sh
Error: Process completed with exit code 1.
```

hernanmd:
I think the error is unrelated to the PR, but the CI is showing "In progress..."

jecisc (approved):

Issue #1493, opened 2023-12-15
SpChooserPresenter keeps filter text after move, yet does not apply it


JanBliznicenko:
If you fill in any filter, then move items, then the filter text keeps showing, yet the filter is no longer applied.

```
	| targetItemToRemove sourceItemToAdd sourceItems targetItems allItems presenter |
	targetItemToRemove := 'TargetToRemove'.
	sourceItemToAdd := 'SourceToAdd'.
	sourceItems := {
		               'S1'.
		               'S2'.
		               sourceItemToAdd }.
	targetItems := {
		               'T1'.
		               'T2'.
		               targetItemToRemove }.
	allItems := sourceItems , targetItems.

	presenter := SpChooserPresenter
		             sourceItems: sourceItems copy
		             chosenItems: targetItems copy.

	presenter open

	presenter sourceList filterInputPresenter text: 'ToAdd'.
	presenter targetList filterInputPresenter text: 'ToRemove'.
	presenter removeAll.
```
![Example](https://github.com/pharo-spec/Spec/assets/9571633/cd615b18-1cce-4f77-aa24-0472ce259faa)

Possible solutions:
1. reapply filters after move
1. clear the filter texts after move

Clearing the text might be simpler and user might probably want to rest those filter anyway after a move is done.

JanBliznicenko:
Fixed by https://github.com/pharo-spec/Spec/pull/1494

PR #1494, opened 2023-12-15
Fixed ChooserPresenter moving and filtering items

### LINKED ISSUE #1340: SpChooserPresenter swallows items after removing them and then adding others
**Bug description**
After opening an instance of SpChooserPresenter and opening it as a window and then removing all or a item(s) from the targetList via a removal button (buttons named '<' or '<<') and then selecting an item in the sourceList and adding it via the button named '>' to the targetList removes/hides all previously removed items from the sourceList and doesn't allow to select or add them anymore.

**To Reproduce**
1. Open an SpChooserPresenter in the playground with e.g. `(SpChooserPresenter sourceItems: #('Apple' 'Pear' 'Melon') chosenItems: #('Peach')) open.`
2. Click on button '<<' or alternatively select 'Peach' and then click on button '<'
3. Now select for example 'Melon' and then click on the button '>'

Now 'Melon' is shown in the targetList but 'Peach' has disappeared from the sourceList is not shown there anymore, thus preventing selecting and adding it again if desired.

**Expected behavior**
After removing an item or multiple items from the targetList via a button and then adding other items to targetList, the previously removed items should stay selectable and addable in the sourceList.

**Screenshots**
Opened SpChooserPresenter:
chooser_before

After moving all or one item to sourceList and selecting a new item for moving to targetList:
chooser_between

After moving previously selected item to targetItems list:
chooser_after

Peach has disappeared from sourceList.

**Version information:**
 - OS: Slackware 15.0 64bit
 - Pharo 11 latest

**Expected development cost**
I don't know yet, but I think it must be something in the methods `#removeAll` or `#removeSelected`.  At first glance both look fine to me, but somehow the items disappear. I haven't investigated much yet as I hadn't time.

**Additional context**
None so far

JanBliznicenko:
Fixes https://github.com/pharo-spec/Spec/issues/1340, https://github.com/pharo-project/pharo/issues/13907 and https://github.com/pharo-spec/Spec/issues/1493.
* In order to fix losing items, instead of remembering original sourceItems then trying to re-calculate from these after every move, use actual items in both lists
* Merged duplicite parts of code for moving items
* In order to fix filtering after moves: If the filter would result in empty list, clear the filter. if the filter would result in at least some shown items, reapply the filter. Always match filter field text with actually applied filter.
* Moved sorting responsibility to ListPresenters (use SpAbstractListPresenter sortingBlock) instead of Chooser replacing lists items after sorting "manually" (the chooser now only provides sortingBlock during initialization)

### DIFF
--- a/src/Spec2-CommonWidgets-Tests/SpFilteringListPresenterTest.class.st
+++ b/src/Spec2-CommonWidgets-Tests/SpFilteringListPresenterTest.class.st
@@ -7,8 +7,9 @@ Class {
 	#instVars : [
 		'listWithFilter'
 	],
-	#category : 'Spec2-CommonWidgets-Tests',
-	#package : 'Spec2-CommonWidgets-Tests'
+	#category : 'Spec2-CommonWidgets-Tests-Core',
+	#package : 'Spec2-CommonWidgets-Tests',
+	#tag : 'Core'
 }
 
 { #category : 'accessing' }
@@ -76,7 +77,7 @@ SpFilteringListPresenterTest >> testEnableItemSubstringFilter [
 SpFilteringListPresenterTest >> testFilterListItems [
 
 	| listItems |
-	listItems := { 
+	listItems := {
 		             OrderedCollection.
 		             Array.
 		             SequenceableCollection.
@@ -84,7 +85,7 @@ SpFilteringListPresenterTest >> testFilterListItems [
 		             Dictionary }.
 	listWithFilter items: listItems.
 	listWithFilter filterListItems: 'collection'.
-	self assertCollection: listWithFilter items hasSameElements: { 
+	self assertCollection: listWithFilter items hasSameElements: {
 			OrderedCollection.
 			SequenceableCollection }.
 	listWithFilter filterListItems: 'xyz'.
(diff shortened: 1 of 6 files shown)


src/Spec2-CommonWidgets-Tests/SpFilteringListPresenterTest.class.st
```diff
@@ -7,8 +7,9 @@ Class {
 	#instVars : [
 		'listWithFilter'
 	],
-	#category : 'Spec2-CommonWidgets-Tests',
-	#package : 'Spec2-CommonWidgets-Tests'
+	#category : 'Spec2-CommonWidgets-Tests-Core',
+	#package : 'Spec2-CommonWidgets-Tests',
+	#tag : 'Core'
```
estebanlm, line 12: convention is actually put tag before package (because then the template will work correctly, otherwise it doesn't). I do not think this is an issue in monticello, however... maybe @jecisc knows ?

estebanlm (commented):
let's wait for what Cyril has to say, otherwise it looks ok, I will approve preemptively :P

src/Spec2-CommonWidgets-Tests/SpFilteringListPresenterTest.class.st
jecisc: I added the tag after the package in the Tonel format because I did not really think about the order in other places :( 

So this is not related to this PR

PR #1496, opened 2023-12-22
Defer tests to the morphic ui thread

guillep:
Make Spec tests run in the UI thread to avoid concurrency issues.

The problem is more detailed in here: https://github.com/pharo-project/pharo/issues/12502

Problem summary: during the test run in the CI, tests and morphic run in separate tests.
Thus, a test might change a graphic element during the morphic ui draw and break it.
If an exception happens in a thread other than the test thread, the process quits and stops the test run with an unexpected outcome.

### DIFF
--- a/src/Spec2-Code-Tests/SpCodeInteractionModelTest.class.st
+++ b/src/Spec2-Code-Tests/SpCodeInteractionModelTest.class.st
@@ -1,6 +1,6 @@
 Class {
 	#name : 'SpCodeInteractionModelTest',
-	#superclass : 'TestCase',
+	#superclass : 'SpBaseTest',
 	#instVars : [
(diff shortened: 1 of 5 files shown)


guillep:
There is a broken test, unrelated, and caused by some protocol recompilation.

To see the issue in isolation of this PR, the following code adds a new superclass and breaks the test:
```smalltalk
ProperPackagesTest debug: #testPackageExtensionsStartsWithProperPackageName.

newTestCase := (TestCase << #NewTestCase)
	package: #toto;
	install.
	
(newTestCase << #SpBaseTest)
	install.
	
ProperPackagesTest debug: #testPackageExtensionsStartsWithProperPackageName.
```

PR #1500, opened 2023-12-27
introduce displayBlock and deprecate display.

Ducasse:
This is a block and not a display that we get!

### DIFF
--- a/src/Spec2-CommonWidgets/SpChooserPresenter.class.st
+++ b/src/Spec2-CommonWidgets/SpChooserPresenter.class.st
@@ -211,7 +211,7 @@ SpChooserPresenter >> initializePresenters [
 SpChooserPresenter >> initializeSortingBlocks [
 
 	| sortingBlock |
-	sortingBlock := sourceList display ascending.
+	sortingBlock := sourceList displayBlock ascending.
 	sourceList sortingBlock: sortingBlock.
 	targetList sortingBlock: sortingBlock
 ]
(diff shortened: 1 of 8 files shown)


estebanlm:
Stef, please don't. 
This is not a good change . 
Some years ago we decided to drop the block suffix in purpose and favour just "display:". The rationale is that it makes a whole family of accessors play better together, and also... is not just a "block" what you should pass there but any valuable. 

This changes makes everything more dirty.

estebanlm:
I will revert this PR

estebanlm:
> Stef, please don't. This is not a good change . Some years ago we decided to drop the block suffix in purpose and favour just "display:". The rationale is that it makes a whole family of accessors play better together, and also... is not just a "block" what you should pass there but any valuable.
> 
> This changes makes everything more dirty.

also, the idea is that you will set what is going to be display. We don't care the nature of the parameter :)

Ducasse:
Your API is not good. 
`display` is not a good API,  `display:` is not a good API. 
But you do not listen to your users, else you would ask them. 

And display is an order and it conflicts with the display. 
You see at the end I do not care but this has also consequences.

Ducasse:
And BTW I do not see any evidence about 

```
he rationale is that it makes a whole family of accessors play better together, and also... is not just a "block" what you should pass there but any valuable.
````

We can always pass a valuable instead of a block so this is a non argument.

**Comment by Ducasse** [comment]
With this kind of design it means that we have to constantly check the code to be able to pass an argument. It looks like glamour APIs. Always guessing. Now I do not care but I will not document this .

**Comment by Ducasse** [comment]
You see this design

```
displayUnderline: aBlock
	"If `aBlock` answer true, the cell will be displayed underlined.
	 `aBlock` receives one argument (the model element)."

	underlineAction := aBlock
```

prevent you to have 


displayUnderline: true
and displayUnderlineBlock: [:item | item isxxx ]

Now you force people to guess that this is a block and to always pass a block when true is just enough.

Ducasse:
So what is the argument of `displayIcon:` ... this is not an icon! too bad.

PR #1504, opened 2024-01-11
Revert "introduce displayBlock and deprecate display."

estebanlm:
Reverts pharo-spec/Spec#1500


guillep:
Hi it's important to revert this ASAP. Nothing works (debugger, iceberg...) in Pharo when there is a deprecation raised on the UI.

PR #1511, opened 2024-01-23
adding guard in SpMorphicMillerAdapter>>#scrollToLastPage to prevent SubscriptOutOfBounds

adri09070:
Fixes partially https://github.com/pharo-spec/NewTools/issues/664

See discussion here: https://github.com/pharo-spec/NewTools/pull/665

### DIFF
--- a/src/Spec2-Adapters-Morphic/SpMorphicMillerAdapter.class.st
+++ b/src/Spec2-Adapters-Morphic/SpMorphicMillerAdapter.class.st
@@ -236,8 +236,9 @@ SpMorphicMillerAdapter >> scrollToShowLastPage [
 	self widget defer: [
 		| size firstMorphIndex |
 		size := innerWidget submorphs size.
-		firstMorphIndex := size - self presenter visiblePages + 1.
+
 		size > 0 ifTrue: [
+			firstMorphIndex := size - self presenter visiblePages + 1.
 			self widget showMorph:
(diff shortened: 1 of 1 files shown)


src/Spec2-Adapters-Morphic/SpMorphicMillerAdapter.class.st
```diff
@@ -233,10 +233,13 @@ SpMorphicMillerAdapter >> scrollByDeltaPercent: percent [
 { #category : 'private' }
 SpMorphicMillerAdapter >> scrollToShowLastPage [
 
-	self widget defer: [ | size firstMorphIndex |
+	self widget defer: [
+		| size firstMorphIndex |
 		size := innerWidget submorphs size.
-		firstMorphIndex := size - self presenter visiblePages + 1. 
-		self widget showMorph: (innerWidget submorphs at: ((firstMorphIndex max: 1) min: size)) ]
+		firstMorphIndex := size - self presenter visiblePages + 1.
```
estebanlm: I would put this line inside the guard... no need on calculate the index if we are not going to use it.

estebanlm (changes requested):

src/Spec2-Adapters-Morphic/SpMorphicMillerAdapter.class.st
adri09070: done

PR #1507, opened 2024-01-21
Set the minWidth and the minHeight of the endPanel created for a box layout

koendehondt:
See issue https://github.com/pharo-project/pharo/issues/15830. As described there, this PR:
* Sets the `minWidth` of an `endPanel` in a horizontal box layout
*  Sets the `minHeight` of an `endPanel` in a vertical box layout

Please apply this fix to Pharo 11 as well. Thank you!

PR #1513, opened 2024-01-26
Set the minWidth and the minHeight of the endPanel created for a box layout

koendehondt:
@estebanlm 

The PR for Pharo 12 was https://github.com/pharo-spec/Spec/pull/1507.

This is the PR to fix the same problem in Pharo 11.

### DIFF
--- a/src/Spec2-Adapters-Morphic/SpMorphicBoxAdapter.class.st
+++ b/src/Spec2-Adapters-Morphic/SpMorphicBoxAdapter.class.st
@@ -212,6 +212,7 @@ SpMorphicBoxAdapter >> newHorizontal [
 			listDirection: #leftToRight;
 			hResizing: #shrinkWrap;
 			width: 0;
+			minWidth: 0; "Otherwise an empty endPanel has a width of 2 pixels due to Morph>>#minWidth."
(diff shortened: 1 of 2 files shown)


koendehondt:
The changes made two box layout tests fail, because of rounding errors on the width of the children of the box layout. I adapted the code to avoid rounding errors.
On CI, the initial window width is 240 pixels instead of the 400 pixels in `SpWindowPresenter>>#defaultInitialExtent`. That made the tests fail on CI, while they were green locally. That issue accounts for the many commits to adapt the tests.
The tests used a list, but its initial width is bigger than the space allotted by the box layout. That is why some assertions used a `>=` instead of `=`. I replaced the list by a button.
I wonder whether the previous issue was the reason to add `#skipOnPharoCITestingEnvironment` in two tests. I took them out, as the tests are green on CI now.

Btw: does `#skipOnPharoCITestingEnvironment` actually avoid running a test on CI? They were run and they failed, so it is probably a good idea to check whether the environment variable is in use on CI.

koendehondt:
@estebanlm The tests are green except for one that does not seem to be related to the changes. Please review and merge.

PR #1515, opened 2024-01-28
Cleanup/remove rename default spec to default layout refactoring

Ducasse:


estebanlm:
seems that clean up is not letting the package really empty ?

Ducasse:
I do not know. 
We will check.

Ducasse:
Now the PR just remove one package so I do not know how it could stay and be not empty.

Ducasse:
@balsa-sarenac I have the impression that you removed too much. 
Why do we remove the full package? 
I would not remove 
```
package: 'Spec2-Microdown' with: [ spec requires: #('SpecCore' 'Spec2-Adapters-Morphic') ];
```
Can you have a look?

balsa-sarenac:
@Ducasse The package was empty after I have removed the class so I removed it as well.
Packages that are left are: `#('SpecCore' 'Spec2-Adapters-Morphic')`
I have only removed `#('Spec2-Transformations')`

Ducasse:
Ok. Because on my machine it contains a lot of extensions.

jecisc:
Should this be merged?

estebanlm:
yeah, errors not related

PR #1533, opened 2024-04-05
Improve SpSelectDialog for handling line splitting texts

### LINKED ISSUE #16403: Refactoring options popup does not have enough space for the message
Remove a class with users:

imagen

the message is cut:

imagen

tesonep:
Fix https://github.com/pharo-project/pharo/issues/16403

- Adding style for disabled text presenter
- Adding extension method to text to calculate line heights with wrapping.
- SpSelectDialog layout the label using a text taking in account the wrapping and height.


demarey:
very nice! Thank you @tesonep

Ducasse:
Oh this is super nice. 
may be the logic of wrapping should be pushed in the superclass so that all dialogs get it.

PR #1535, opened 2024-04-09
Fix selection bug in SpFilteringSelectableListPresenter

jecisc:
SpFilteringSelectableListPresenter has items and a list of the selected items. But if we change the items while a selection exists, we end up with selected elements that are not in the list. 

I propose to filter out those selected elements when we update the items. This fixes a bug in the SpChooserPresenter when we could have duplicated elements in the target list if we clicked multiple times on the "move elements" button.


jecisc:
Failing tests are not related

ClotildeToullec (approved):

PR #1545, opened 2024-05-18
Remove deprecated code

jecisc:
This change removes all the deprecated classes and methods to start clean for P13


Ducasse:
@koendehondt this makes me think that we should base the book on P13

koendehondt:
> @koendehondt this makes me think that we should base the book on P13

I don't see it this way. First, fixing the Pharo version to 12 for the book is one variable less. Second, `activatesOnDoubleClick` and `SpApplication>>#close` are not used in the code examples.

PR #1548, opened 2024-05-28
New dialog to select multiple items using checkboxes

carolahp:
The new dialog is displayed below. 
Its api is the same as SpSelectDialog. 
Three examples were added.

image

This presenter is used in Pharo [PR#16637](https://github.com/pharo-project/pharo/pull/16637)

### DIFF
--- a/src/Spec2-Dialogs/SpApplication.extension.st
+++ b/src/Spec2-Dialogs/SpApplication.extension.st
@@ -109,6 +109,12 @@ SpApplication >> newSelect [
 ]
 
 { #category : '*Spec2-Dialogs' }
+SpApplication >> newSelectMultiple [
+
+	^ SpSelectMultipleDialog newApplication: self
+]
+
+{ #category : '*Spec2-Dialogs' }
 SpApplication >> notificationClass [ 
 
 	^ SpNotificationItem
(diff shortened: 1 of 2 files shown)


demarey:
This presenter is nice but it looks to be a duplicate of `SpChooserPresenter`.
The new dialog also misses a filter for items.

carolahp:
@demarey I think the `SpChooserPresenter` is designed to selecting items from a long lists (as in the image), while the presenter I propose works best for a short list of items.
image

You are right that my presenter is missing a filter. I can add it

demarey:
Hi @carolahp 
I agree your presenter is better for a few items but I still think the same presenter can do both jobs:
- present a logn list of items to select: `SpChooserPresenter`
- present a short list of items to select: `SpSelectMultipleDialog`
I think `SpChooserPresenter` could have 2 different layouts for both usage:
- one showing twol list,
- the other only showing one list.
Could you have a look at `SpChooserPresenter` to see if it is doable. The only *hard* point I can see is the different behaviour when selecting items from the first list but maybe I miss something.

carolahp:
Hi @demarey 
Modifying `SpChoosedPresenter` to do both jobs would implies:
- Modifying the logic behind the way `chosenItems` are computed and displayed. When the items list is short the chosen items are the checked items in the first list, but when the items list is long the chosen items are those in the second list. I can't think of a clean way to introduce this distinction in the code, but to add many `if`s 
Maybe it's just me who can't find an elegant way to implement this new behavior to `SpChooserPresenter`. Could you please give me your opinion?

PR #1549, opened 2024-05-29
Fix `SpCodePresenter>>#evaluate:onCompileError:onError:` that does not consider interaction model's bindings when evaluating code

### LINKED ISSUE #1506: `SpCodePresenter>>#evaluate:onCompileError:onError:` does not take into account additional bindings when evaluating an expression
When evaluating an expression, the code presenter does not consider the bindings of its interaction model:

```Smalltalk
SpCodePresenter>>#evaluate:onCompileError:onError:
    result := self interactionModel compiler
		            source: aString;
		            environment: self environment;
		            failBlock: [
			            self announcer announce:
					            (SpCodeEvaluationFailedAnnouncement newContent:
							             aString).
			            ^ compileErrorBlock value ];
		            evaluate.
```

So, if an interaction model allows to add additional bindings, **it leads to a popup because the variable is not recognized, even if the binding DOES exist in the interaction model:**

image

**The problem appears because**, in a code presenter, `SpCodeSelectionCommand>>#evaluate:andDo:` calls `SpCodePresenter>>#evaluate:onCompileError:onError:` which does not take into account additional bindings.

**The problem DOES NOT appear in the playground though, and I can't figure out why because it does use additional bindings ...* One lead that could be tracked is that it does not use the same command class (it overrides the context menu to use `StEvaluateCommand` instead, which still calls `SpCodePresenter>>#evaluate:onCompileError:onError:` in the end...)

**Proposed fix: I think we should just provide the bindings from the interaction model to the compiler:**

```Smalltalk
SpCodePresenter>>#evaluate:onCompileError:onError:
     result := self interactionModel compiler
		            source: aString;
		            environment: self environment;
		            bindings: self interactionModel bindings;
		            failBlock: [
			            self announcer announce:
					            (SpCodeEvaluationFailedAnnouncement newContent:
							             aString).
			            ^ compileErrorBlock value ];
		            evaluate.
```

adri09070:
Fixes #1506 

I do not understand why the problem does not appear when evaluating code in the playground for example, but still, it is wrong not to use the bindings of the interaction model when evaluating code

### DIFF
--- a/src/Spec2-Code/SpCodePresenter.class.st
+++ b/src/Spec2-Code/SpCodePresenter.class.st
@@ -425,7 +425,7 @@ SpCodePresenter >> environment: anEnvironment [
 ]
 
 { #category : 'command support' }
-SpCodePresenter >> evaluate: aString onCompileError: compileErrorBlock onError: errorBlock [ 
+SpCodePresenter >> evaluate: aString onCompileError: compileErrorBlock onError: errorBlock [
 	"evaluate aString. 
 	 evaluate compileErrorBlock if there are compilation errors. 
 	 evaluate errorBlock is anything happens *during* evaluation (code compiled, but it does not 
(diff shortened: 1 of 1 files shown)


adri09070:
unrelated failing test

PR #1555, opened 2024-06-13
withWindowDo: should return the result of the block

demarey:
Else, if you would like to trigger the ok action, you will not get back the result of the actions (e.g. selected item)


estebanlm:
indeed

estebanlm:
but then... what it returns if there is no window? still answering `self` sounds weak (I very much prefer `nil` in this case)

demarey:
> but then... what it returns if there is no window? still answering `self` sounds weak (I very much prefer `nil` in this case)

yes, I agree. I will change it to return nil

PR #1565, opened 2024-06-28
merge latest

estebanlm:


estebanlm:
hey @jecisc can you take a look at this?
I can't find any undeclared or package that starts bad in my code so I guess PR is fine.
What *I do find* is a DNU on `testPackageExtensionsStartsWithProperPackageName`, because packageName answer nil in some condition (in some methods that came from traits), and that may be another kind of bug...  
In any case, merging this will unlock the changes on newtools too.

jecisc:
I'm debugging in an image and it seems that we indeed have a bug between traits and protocols.. Protocols should not be nil

jecisc:
This happens only on those 2 methods and I don't know what is special about them...

image

jecisc:
This one is really not easy to debug :/

Issue #1547, opened 2024-05-25
whenWillCloseDo: block is evaluated twice when closing a window


koendehondt:
Context: Pharo 12.

Consider this code:
```smalltalk
| windowPresenter presenter  |
presenter := SpPresenter new.
presenter layout: SpBoxLayout newTopToBottom.
windowPresenter := presenter open.
windowPresenter whenWillCloseDo: [ :announcement |
	(presenter confirm: 'Are you sure that you want to close the window?')
		ifFalse: [ announcement denyClose ] ]
```

Be aware that a `SpWindowWillClose` announcement has a default of allowing a window to close (see `SpWindowWillClose>>#initialize`). That is why the code above does not send `announcement allowClose`.

* Run the code
* Click the close box of the window.
* A confirmation dialog appears.
* Click the "Yes" button in the confirmation dialog.
* (Bug) The confirm dialog is opened again.

Here is the reason:

```smalltalk
SystemWindow>>closeBoxHit
	"The user clicked on the close-box control in the window title.
	For Mac users only, the Mac convention of option-click-on-close-box is obeyed if the mac option key is down.
	If we have a modal child then don't delete.
	Play the close sound now since this is the only time we know that the close is user-initiated."

	self allowedToClose ifFalse: [^self].
	self playCloseSound.
	self close
```
The method `closeBoxHit` sends `close`:

```smalltalk
SpWindow>>close 

	self announceWillClose ifFalse: [ ^ self ].
	super close.
```

`close` sends the announcement:
```smalltalk
SpWindow>>announceWillClose
	| announcement |

	announcement := SpWindowWillClose new
		window: self;
		yourself.
	self announce: announcement.
	self currentWorld announcer announce: announcement.
	^ announcement canClose 
```

That will trigger the `whenWillCloseDo: ` block in the code snippet.

`SpWindow>>close` does a super send, because `self announceWillClose` evaluates to `true` (the user clicked "Yes").

The method in the superclass is:

```smalltalk
SystemWindow>> close
	^ self delete
```
and `delete` is:

```smalltalk
StandardWindow>>delete
	"If fullscreen remove the owner too."

	self mustNotClose ifTrue: [^ self].
	self model ifNotNil: [
		self model okToChange ifFalse: [ ^ self ].
		self model okToClose ifFalse: [ ^ self ] ].
	self isFullscreen
		ifTrue: [self owner delete]
		ifFalse: [super delete]
```

It does super send on the last line:

```smalltalk
[...]

PR #1566, opened 2024-06-28
Fix: #1547: whenWillCloseDo: block is evaluated twice when closing a window 

Ducasse:
We decided to remove the closeDiscardingChanges from Spec

### DIFF
--- a/src/Spec2-Adapters-Morphic/SpWindow.class.st
+++ b/src/Spec2-Adapters-Morphic/SpWindow.class.st
@@ -47,13 +47,6 @@ SpWindow >> close [
 ]
 
 { #category : 'open/close' }
-SpWindow >> deleteDiscardingChanges [
-
-	self announceWillClose.
-	^ super deleteDiscardingChanges
-]
-
-{ #category : 'open/close' }
 SpWindow >> extent: aPoint [ 
 	| announcement oldExtent |
 


Ducasse:
This is a potential backport to P12.

estebanlm:
please do not merge this, I let a comment in the reported issue.

MarcusDenker:
What is important, too: please add a better name tan "Fix: #1547" (this looks very bad in the weekly email, too)

Ducasse:
Right now the code of the book breaks.

estebanlm:
> Right now the code of the book breaks.

can you point me to the example that breaks ?

Ducasse:
@koendehondt Do you remember this is in the window chapter?

koendehondt:
> > Right now the code of the book breaks.
> 
> can you point me to the example that breaks ?

The example in the description of https://github.com/pharo-spec/Spec/issues/1547 exposes the bug.

The code in the book is in section 9.3 "Preventing window close". The corresponding code is in class `WindowExamplePresenter2` in the [code repo](https://github.com/SquareBracketAssociates/CodeOfSpec20Book/blob/master/src/CodeOfSpec20Book/WindowExamplePresenter2.class.st). The class comment links to https://github.com/pharo-spec/Spec/issues/1547 to indicate that the example does not work.

estebanlm:
This is fixed upstream (I applied same thing)

koendehondt:
> This is fixed upstream (I applied same thing)

Could you post the commit here? Thanks.

PR #1571, opened 2024-07-16
Revert "Fix dependency problem"

jecisc:
Reverts pharo-spec/Spec#1570


jecisc:
This breaks Pharo because Spec is loaded after the diff morphs

PR #1573, opened 2024-07-20
Enable specification of a comparison strategy to compare list items

hernanmd:
This change allows for custom comparison strategies to `SpCollectionListModel`. The main changes are:

1. Added `comparisonStrategy` instance variable with getter and setter.
2. Changed `indexOf:ifAbsent:` method to use the comparison strategy.
3. Initialized `comparisonStrategy` with identity comparison in the `initialize` method.

### DIFF
--- a/src/Spec2-Core/SpCollectionListModel.class.st
+++ b/src/Spec2-Core/SpCollectionListModel.class.st
@@ -11,7 +11,8 @@ Class {
 	#instVars : [
 		'#announcer',
 		'#collection',
-		'#sorting => ObservableSlot'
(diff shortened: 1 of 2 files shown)


hernanmd:
@jecisc this is what you need to get the CI green again :)

jecisc:
This one has the same problem as a PR done by esteban :( I don't know why but the produced image has methods without protocol which should never happen. And I have a really hard time to debug the problem. 

It's probably not caused by your code directly, but it probably causes something that ends up creating the problem

PR #1577, opened 2024-07-24
Revert "Merge pull request #1565 from pharo-spec/dev-3.0"

jecisc:
This reverts commit 41606bd24c144794c7170c28bce9a56b87104c7a, reversing changes made to 3963b087c0b6855f4059d9c9dc483fc2808548eb.

@estebanlm Just letting you know that I have to revert this because my image becomes unusable after this merge.

For example I get that:

image

Also it broke the dependency tests of Pharo.


estebanlm:
that should not happen unless you create a SpWindow (bad name, by the way) *ad hoc*, since they are just exisitng as part of a presenter (is the morphic representation of a `SpWindowPresenter`  which should be the model element, one must not exist without the other... :(

PR #1578, opened 2024-08-01
Revert "Fix scrollTo: on SpScrollableLayout not working"

jecisc:
Reverts pharo-spec/Spec#1370

This introduces 2 failing tests in Pharo 

image


jecisc:
We're fixing it with Hernan

PR #1586, opened 2024-08-15
Add scroll to index to Lists

hernanmd:
This PR adds a method, `scrollToIndex:`, to support vertical list scrolling in Spec. Currently working in the Morphic backend. Its functionality is necessary to fix multiple issues, such as :

- https://github.com/pharo-spec/NewTools/issues/699 
- https://github.com/pharo-spec/NewTools/issues/808 
- https://github.com/pharo-spec/Spec/issues/1585
- https://github.com/hernanmd/class-diff/issues/1
and probably others

It also adds corresponding tests.

PR #1587, opened 2024-08-16
Adapt test because CI uses another display scale configuration

hernanmd:
The test for the scroll index provided in #1586 was failing because we got a different index number in the CI execution. Simplify the test so now it can now be run interactively and in the CI.


hernanmd:
Failing tests are related to SmalltalkCI still using Author class

PR #1588, opened 2024-08-19
Make implementors of #transferFrom:event: add #shouldCopy from the TransferMorph to the SpDragAndDropTransfer

Rinzwind:
This pull request makes the implementors of `#transferFrom:event:` add `#shouldCopy` from the TransferMorph to the SpDragAndDropTransfer. This is related to [Pharo issue #15233](https://github.com/pharo-project/pharo/issues/15233) for which the ability to hold the Shift key while dragging to indicate that the object should be copied when dropped would seem useful.


Ducasse:
Thanks Kris! 
The build is broken because we should update it (it requests Author that we removes). I will see if I can understand how to change the CI version.

Rinzwind:
OK, I see there’s an issue about that: [smalltalkCI issue #655](https://github.com/hpi-swa/smalltalkCI/issues/655).

Ducasse:
Green bar so I push

PR #1589, opened 2024-08-29
fix faling test Spec2.Code.Tests.SpCodePresenterTest.testDoBrowseImplementors

demarey:


hernanmd:
I cannot assign reviewers but @jecisc or @Ducasse ?

PR #1599, opened 2024-09-10
Fix adapter tests, that should access the widget via the adapter, not directly.

### LINKED ISSUE #1598: Adapter tests access the widget directly instead of doing it via adapter
The idea is to do not talk directly to the widget, because the adapter is who knows how to do it. This is the case for other `*AdapterTest`

tinchodias:
Fixes #1598


Ducasse:
Thanks martin this is a great fix.

Issue #1603, opened 2024-09-15
Bug: Menu item enablement in CmUICommandDisplayStrategy>>#display:in:do: is fixed


koendehondt:
Context : Pharo 12 and Pharo 13.

I think CmUICommandDisplayStrategy>>#display:in:do: is implemented incorrectly. Now it is:

```smalltalk
display: aCmSpecCommand in: aMenuOrGroupPresenter do: aBlock
	aMenuOrGroupPresenter
		addItem: [ :item |
			aBlock value: item.
			item enabled: aCmSpecCommand canBeExecuted.
			item ]
```

As a consequence, a menu item's enablement state is set only once, when the menu item is added. Given that the command decides whether it can be executed, the enablement state should be updated every time a menu item is displayed. In Spec, that is achieved by passing a block to `#enabled:`.

So I think the method should be implemented as:

```smalltalk
display: aCmSpecCommand in: aMenuOrGroupPresenter do: aBlock
	aMenuOrGroupPresenter
		addItem: [ :item |
			aBlock value: item.
			item enabled: [ aCmSpecCommand canBeExecuted ].
			item ]
```

This issue is related to https://github.com/pharo-spec/Spec/issues/1559 and https://github.com/pharo-spec/Spec/issues/1561, which report other enablement issues.

PR #1604, opened 2024-09-15
Use a block to determine the enablement state of a menu item that is created based on a command

koendehondt:
Code changes for https://github.com/pharo-spec/Spec/issues/1603.

I did not add a test, nor an example, because it is not clear to me how to test this. If one can provide me some guidance, I am happy to add the test.


koendehondt:
This change is important for the Spec book. Please merge the PR to Pharo 12 as well.

jecisc:
Hi @koendehondt 
Would it be possible to produce the PR to P12 and ping me on it to merge it?

koendehondt:
> Hi @koendehondt Would it be possible to produce the PR to P12 and ping me on it to merge it?

@jecisc Here it is: https://github.com/pharo-spec/Spec/pull/1605

Thank you!

PR #1608, opened 2024-09-23
New presenter to split sub-presenters

hernanmd:
This PR adds a new composite presenter that contains two presenters. The first presenter is fixed and can be expanded or split. The second presenter is dynamic and can be shown split or hidden. It includes tests for opening with hidden or split view.

## Demo

https://github.com/user-attachments/assets/1d342ee2-b50c-4cbb-b261-7243586c644d

### DIFF
--- a/src/Spec2-CommonWidgets/SpToggleSplitPresenter.class.st
+++ b/src/Spec2-CommonWidgets/SpToggleSplitPresenter.class.st
@@ -1,5 +1,5 @@
 "
-This presenters contains two presenters provided by the user. The first presenter is fixed and can be expanded or splitted. The second presenter is dynamic and can be shown splitted or hidden. Split can be done horizontally or vertically.
+This presenter contains two presenters provided by the user. The first presenter is fixed and can be expanded or splitted. The second presenter is dynamic and can be shown split or hidden. Split can be done horizontally or vertically.
 
 ## Structure
 
@@ -36,7 +36,8 @@ Class {
 		'toggleButton',
 		'showString',
 		'hideString',
-		'layoutDirection'
+		'layoutDirection',
+		'positionOfSlider'
 	],
 	#category : 'Spec2-CommonWidgets-Core',
 	#package : 'Spec2-CommonWidgets',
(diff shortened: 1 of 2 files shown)


Ducasse:
Hi hernan
May be you want to add a splitter and expose an API so that we can preconfigure the proportion and by default keep the equal part for each of the presenter.
Like that we can get it used in many scenario.
S.

Ducasse:
these tests are failing on the server

```
SpToggleSplitPresenterTest
 ✗ #testOpenPresenterIsBuilt (27ms)
 ✗ #testOpenPresenterRaisesDisplayEvent (16ms)
 ✗ #testOpenPresenterRaisesBuiltEvent (30ms)
 ✗ #testExample (41ms)
 ✗ #testOpen (19ms)
 ✗ #testOpenPresenterIsDisplayed (20ms)

ZnClientTest
 ✗ #testGetGeoIP (10141ms)

SystemDependenciesTest
 ✗ #testExternalBasicToolsDependencies (19ms)
 ✗ #testExternalUIDependencies (14ms)
```

src/Spec2-CommonWidgets/SpToggleSplitPresenter.class.st
```diff
@@ -0,0 +1,189 @@
+"
+This presenters contains two presenters provided by the user. The first presenter is fixed and can be expanded or splitted. The second presenter is dynamic and can be shown splitted or hidden. Split can be done horizontally or vertically.
```
Ducasse: This presenter contains.....
and I think that this is split or hidden

hernanmd:
> Hi hernan May be you want to add a splitter and expose an API so that we can preconfigure the proportion and by default keep the equal part for each of the presenter. Like that we can get it used in many scenario. S.

Thanks for the suggestion, I added a `positionOfSlider:` method so users can configure the proportion

Captura de pantalla 2024-09-23 a las 14 56 28

src/Spec2-CommonWidgets/SpToggleSplitPresenter.class.st
hernanmd: Fixed

PR #1610, opened 2024-09-27
Remove senders of scaledByDisplayScaleFactor and displayScaleFactor

guillep:
Remove senders of scaledByDisplayScaleFactor and displayScaleFactor.

This is superseeded by new scale factor support in Pharo12 and on


Ducasse:
Broken tests are related to Refactorings and not to this issue. 
So I integrate them.

PR #1611, opened 2024-09-27
Add SpMenuPresenter>>#fillWith: that applies for SpMenuBarPresenter as well, and add tests and example

koendehondt:
For Pharo 12.

PR for https://github.com/pharo-spec/Spec/issues/1601

* Add SpMenuPresenter>>#fillWith: that applies for SpMenuBarPresenter as well, and add tests and example
* Fix bug: add SpMenuBarPresenterTest>>#classToTest, otherwise the tests use SpMenuPresenter as the class to test


koendehondt:
@estebanlm As requested, here is the PR for Pharo 12. There are 2 failing tests, but they are not related to the changes in this PR.

PR #1612, opened 2024-09-30
Add a simple code file editor so that we can use it for the preferences.

Ducasse:
It replaces the UIManager default edit:  So we can remove this edit: method too.

### DIFF
--- a/src/Spec2-CommonWidgets/SpSimpleFileEditPresenter.class.st
+++ b/src/Spec2-CommonWidgets/SpSimpleFileEditPresenter.class.st
@@ -1,0 +1,142 @@
+"
+I define a simple code editor. 
+Before saving I create a backup of the edited file. 
+
+
+Here is a typical use: 
+
+```
+SpSimpleFileEditPresenter new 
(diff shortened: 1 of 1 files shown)


jecisc (approved):

src/Spec2-CommonWidgets/SpSimpleFileEditPresenter.class.st
```diff
@@ -0,0 +1,142 @@
...
+
+{ #category : 'layout' }
+SpSimpleFileEditPresenter >> defaultLayout [
+
+	^ SpBoxLayout newTopToBottom
+		spacing: 5;
+	  	add: (SpBoxLayout newLeftToRight
+				add: label expand: true;
+				yourself)
+	  		height: self calculateLabelHeight;
```
demarey, line 63: it is not recommanded to hardcode the height in Spec.
Using expand: false on the TopToBottom layout should be enough.
```smalltalk
^ SpBoxLayout newTopToBottom
		spacing: 5;
	  	add: (SpBoxLayout newLeftToRight
				add: label expand: true;
				yourself)
	  		expand: false;
	add: codeText;
	yourself
```

PR #1629, opened 2024-10-18
Setter of unfilteredItems

PalumboN:
Adding a setter for NewTools

### DIFF
--- a/src/Spec2-CommonWidgets/SpFilteringListPresenter.class.st
+++ b/src/Spec2-CommonWidgets/SpFilteringListPresenter.class.st
@@ -265,6 +265,12 @@ SpFilteringListPresenter >> unfilteredItems [
 	^ unfilteredItems
 ]
 
+{ #category : 'accessing' }
+SpFilteringListPresenter >> unfilteredItems: col [
+
+	unfilteredItems := col
+]
+
 { #category : 'api - events' }
 SpFilteringListPresenter >> whenSelectionChangedDo: aBlock [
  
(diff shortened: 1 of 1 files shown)


src/Spec2-CommonWidgets/SpFilteringListPresenter.class.st
```diff
@@ -265,6 +265,12 @@ SpFilteringListPresenter >> unfilteredItems [
 	^ unfilteredItems
 ]
 
+{ #category : 'accessing' }
+SpFilteringListPresenter >> unfilteredItems: col [
```
demarey: `col` is not a meaningful parameter name.
could be `someItems`

src/Spec2-CommonWidgets/SpFilteringListPresenter.class.st
```diff
@@ -265,6 +265,12 @@ SpFilteringListPresenter >> unfilteredItems [
...
+
+	unfilteredItems := col
```
PalumboN:
```suggestion
SpFilteringListPresenter >> unfilteredItems: someItems [

	unfilteredItems := someItems
```

src/Spec2-CommonWidgets/SpFilteringListPresenter.class.st
PalumboN: Agreeee!!! Changed! 💯

demarey (approved):

PR #1646, opened 2024-11-10
SpPaginatorPageMorph and SpPaginatorSelectionMorph fixes for theme

### LINKED ISSUE #1645: SpPaginatorPageMorph>>newContentMorph should use theme background
I started to fill an issue on pharo tracker, but I guess this should get here instead.

Working with dark themes, I was bothered by the inspector. When several pages are opened, small white blocks show the offscreen pages. I looked at UITheme and subclasses to see if I could change that, but that’s baked in the code.

The initialization method already accesses the theme for the border color. Instead of `Color white` the method should use the theme background color.

terpon:
replace hard-coded white color with theme background color: fixes #1645


terpon:
Related change is you cannot change #paginationSelectionColor by implementing it in your ColorConfigurator class

PR #1656, opened 2024-11-17
less extension.

Ducasse:
We should minimize extension use!


Ducasse:
The broken tests are unrelated. Reflectivity and SettingBrowser

PR #1671, opened 2024-11-26
1670 introduce a common st command superclass

Ducasse:
- Sorting command into NewTools and 
- other default presenters


estebanlm:
you cannot do an "St" (which would be newtools dependent) superclass in spec. There is a layer violation there.

Ducasse:
But after we will move the commands St to NewTools. 
 Keeping the current names is not nice. Now you can drop the PR because I lost time for this and I'm not in the mood to lose more time.

It loosk in Pharo we put extra luxury to create rules that force us to spend more time on not exciting and interesting issues.

PR #1678, opened 2024-11-29
add drag transfer customization

estebanlm:


estebanlm:
I'm also sneaking an unrelated bugfix around action system, sorry for that :P

PR #1679, opened 2024-11-29
Make #transferFor:from: on SpMorphicTreeTableDataSource delegate to the adapter

Rinzwind:
This pull request makes `#transferFor:from:` on SpMorphicTreeTableDataSource delegate to the adapter as is done in the corresponding method on SpMorphicListDataSource.


Rinzwind:
@estebanlm Could you maybe review this?

PR #1680, opened 2024-12-03
Make #recalculatePageWidths and #recalculatePageHeights on SpMorphicMillerAdapter set the coordinates and dimensions to integral numbers

### LINKED ISSUE #17384: Buttons in inspector move when moving the hand over them
As demonstrated in the following video, some of the buttons in inspector windows move when moving the hand over them. The Playground snippet used in the video shows the position of the inspector’s right-hand object-centric debug points button on the transcript. The position is 516.5@13.0 initially, then changes five times as the hand is moved over it. Various other morphs, including those for the names and values of the variables in the table below it, move position along with the button as well. An earlier, similar issue is issue #16389.

https://github.com/user-attachments/assets/411ae6a4-3eb1-40bc-adc1-390dc32a98cb

Rinzwind:
This pull request makes #recalculatePageWidths and #recalculatePageHeights on SpMorphicMillerAdapter set the coordinates and dimensions to integral numbers which fixes [Pharo issue #17384](https://github.com/pharo-project/pharo/issues/17384). Note that the widgets will no longer necessarily all have the same width or height, they may differ by 1.

### DIFF
--- a/src/Spec2-Adapters-Morphic/SpMorphicMillerAdapter.class.st
+++ b/src/Spec2-Adapters-Morphic/SpMorphicMillerAdapter.class.st
@@ -155,17 +155,19 @@ SpMorphicMillerAdapter >> newVertical [
 { #category : 'private' }
 SpMorphicMillerAdapter >> recalculatePageHeights [
 
-	| newHeight widgets height |
+	| newHeight widgets height bottom |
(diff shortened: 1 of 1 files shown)


Rinzwind:
@estebanlm Could you maybe review this (too)?

PR #1681, opened 2024-12-13
fix search of element in SpCollectionListModel.

demarey:
It should use equality. Bug introduced in commit 3967790c4be27aab48d6c61596f90f4a5ca201a9.
It has already been fixed in Pharo 13. see https://github.com/pharo-spec/Spec/commit/1b48d5a8fd20b88223b635c1620c7f2c4ec3b9f1 and https://github.com/pharo-spec/Spec/commit/23bc559e5f43954409fabda30c89daaaf6f18c6d


demarey:
Fixes #1683

estebanlm (approved):
but you need to dix the iceberg problem yet ;)

demarey:
> but you need to dix the iceberg problem yet ;)

It is. don't you see the fix in `SpDropListItem` comparison?

PR #1699, opened 2025-01-14
Remove duplicated methods

jecisc:
from their superclasses

### DIFF
--- a/src/Spec2-Core/SpWindowPresenter.class.st
+++ b/src/Spec2-Core/SpWindowPresenter.class.st
@@ -578,6 +578,11 @@ SpWindowPresenter >> taskbarIcon [
 ]
 
 { #category : 'private' }
+SpWindowPresenter >> title [
+	^ titleHolder
+]
+
+{ #category : 'private' }
 SpWindowPresenter >> title: aString [
 	titleHolder := aString
 ]
@@ -737,6 +742,11 @@ SpWindowPresenter >> windowClosed [
 	
 	isClosed := true.
 	self application windowClosed: self
+]
+
+{ #category : 'private' }
+SpWindowPresenter >> windowIcon [
+	^ windowIcon
 ]
 
 { #category : 'private' }
--- a/src/Spec2-Core/SpWindowPresenter.class.st
+++ b/src/Spec2-Core/SpWindowPresenter.class.st
@@ -578,11 +578,6 @@ SpWindowPresenter >> taskbarIcon [
 ]
 
 { #category : 'private' }
-SpWindowPresenter >> title [
-	^ titleHolder
-]
-
-{ #category : 'private' }
 SpWindowPresenter >> title: aString [
 	titleHolder := aString
 ]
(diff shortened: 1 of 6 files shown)


src/Spec2-Core/SpWindowPresenter.class.st
```diff
@@ -744,11 +739,6 @@ SpWindowPresenter >> windowClosed [
 	self application windowClosed: self
 ]
 
-{ #category : 'private' }
-SpWindowPresenter >> windowIcon [
```
demarey: should not be removed because parent is marked as to be removed

src/Spec2-Core/SpWindowPresenter.class.st
```diff
@@ -577,11 +577,6 @@ SpWindowPresenter >> taskbarIcon [
 				ifNil: [ self taskbarIconName ]) ]
 ]
 
-{ #category : 'private' }
-SpWindowPresenter >> title [
```
demarey: should not be removed because parent is marked as *DO NOT USE*

src/Spec2-Core/SpWindowPresenter.class.st
jecisc: I reverted it, thanks!

src/Spec2-Dialogs/SpSelectDialog.class.st
```diff
@@ -110,18 +110,6 @@ SpSelectDialog >> displayIcon: aBlock [
 	list displayIcon: aBlock
 ]
 
-{ #category : 'api' }
-SpSelectDialog >> extent [
```
demarey, line 114: should not be removed because parent is marked as to be removed

src/Spec2-Dialogs/SpSelectDialog.class.st
```diff
@@ -110,18 +110,6 @@ SpSelectDialog >> displayIcon: aBlock [
...
-
-	^ initialExtent ifNil: [ self class defaultExtent ]
-]
-
-{ #category : 'api' }
-SpSelectDialog >> extent: aPoint [
```
demarey, line 120: should not be removed because parent is marked as to be removed

src/Spec2-Dialogs/SpSelectMultipleDialog.class.st
```diff
@@ -136,18 +136,6 @@ SpSelectMultipleDialog >> displayIcon: aBlock [
 	self refreshTable
 ]
 
-{ #category : 'api' }
-SpSelectMultipleDialog >> extent [
```
demarey, line 140: should not be removed because parent is marked as to be removed

src/Spec2-Dialogs/SpSelectMultipleDialog.class.st
```diff
@@ -136,18 +136,6 @@ SpSelectMultipleDialog >> displayIcon: aBlock [
...
-
-	^ initialExtent ifNil: [ self class defaultExtent ]
-]
-
-{ #category : 'api' }
-SpSelectMultipleDialog >> extent: aPoint [
```
demarey, line 146: should not be removed because parent is marked as to be removed

demarey (changes requested):

Ducasse:
What we see is that the API is not clean. 
We faced many issues with Alexis such as initialExtent should not be defined but subclasses do it. We discussed with esteban and we will clean this using preferredExtent.
To have only one way to do things.
When cruft accumulate not getting over is a way to block us in the future.
Now it would be nice that this is not the people with the least knowledge that do it.

PR #1715, opened 2025-02-07
Fixes # 17774 on Pharo

AlexisCnockaert:
@jecisc

### DIFF
--- a/src/Spec2-Dialogs/SpRequestDialog.class.st
+++ b/src/Spec2-Dialogs/SpRequestDialog.class.st
@@ -16,7 +16,7 @@ Class {
 { #category : 'accessing' }
 SpRequestDialog class >> defaultPreferredExtent [
 
-	^ 600@200
+	^ 400@200
 ]
 
 { #category : 'documentation' }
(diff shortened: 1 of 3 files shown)


src/Spec2-Dialogs/SpRequestDialog.class.st
```diff
@@ -168,9 +168,15 @@ SpRequestDialog >> initializePresenters [
 
 { #category : 'initialization' }
 SpRequestDialog >> initializeWindow: aWindowPresenter [
-	
+
 	super initializeWindow: aWindowPresenter.
-	aWindowPresenter whenOpenedDo: [ textInput takeKeyboardFocus; selectAll  ]
+	self preferredExtent: 400@200.
```
demarey: why not doing `self preferredExtent: self class defaultPreferredExtent`?
I also wonder if defaultPreferredExtent is used by default.

src/Spec2-Dialogs/SpRequestDialog.class.st
```diff
@@ -14,9 +14,9 @@ Class {
 }
 
 { #category : 'accessing' }
-SpRequestDialog class >> defaultExtent [
+SpRequestDialog class >> defaultPreferredExtent [
 
-	^ 450@170
+	^ 400@200
```
demarey, line 19: Why changing the default size?

PR #1738, opened 2025-03-01
17310 spec UI demo date

### LINKED ISSUE #17310: Incomplete Form example for SpecUI Demo
In Pharo 12 or 13 go to "Help" -> "Spec2 Demo". The first demo is about a form and it shows
how to use edit fields (Name, Surname), number input and so on. 

One can submit the result which is display on the left.

But the result on the left also includes a **date** and a (list) **items** result which is missing
a data picker and a list on the right form:

image

So this example seems incomplete work

- we should either remove date and items from the result
- or the form should demo also how to use a datepicker (SpDatePresenter) and lists, tables or comboboxes (SpListPresenter, ...) easily

pavel-krivanek:
fixes https://github.com/pharo-project/pharo/issues/17310


Ducasse:
Thanks Pavel!

Issue #1732, opened 2025-02-28
Remembering extent should only be done on resizing


Ducasse:
Right now if we set the extent the defaultPreferredSize is not used again
so this prevents having dynamic computation of the extend like a simple conditional. 

```
preferredExtent
	
	^ customExtent ifNil: [ customExtent := self defaultPreferredExtent ]
```

My impression is that this assignment should only happen when invoked from the 

```
initializeWindow: aWindowPresenter
	"override this to set window values before opening. 
	 You may want to add a menu, a toolbar or a statusbar"

	"IMPORTANT: Please ovirride this method and set yourself the informations you want in your window.
	The content of this method is here to help the transition between Spec 1 and 2.
	In the next Spec version the content of this method will be removed and it will do nothing by default because the goal is to remove the management of all of those informations from Composable to put them in WindowPresenter."

	self hasInitialPosition ifTrue: [ aWindowPresenter initialPosition: self initialPosition ].
	aWindowPresenter
		title: self title;
		initialExtent: (self extent ifNil: [self preferredExtent]);
		windowIcon: self windowIcon;
		whenResizingDo: [ :ann | self preferredExtent: ann newSize ].
```

PR #1739, opened 2025-03-02
Fixes: #1732

Ducasse:
- This is only when the change of extent is done via resize that it should be remembered. Else the extent is dynamic (not stored).
- Added nice tests covering this behavior.


Ducasse:
The broken tests are not related.

PR #1741, opened 2025-03-03
Add nil guard to fix random debugger

jecisc:
It happens to me that I get a debugger while closing some windows programatically so I'm proposing to add a nil guard to not have the problem


Ducasse:
Really strange but let us protect us.

PR #1743, opened 2025-03-07
New getter list for SpSelectDialog

AlexisCnockaert:
- Needed it to be able to apply events using the items of the list in the dialog


AlexisCnockaert:
@Ducasse

PR #1747, opened 2025-03-15
introduce an API to be able to reset the last rememberedSize of a pre…

Ducasse:
…senter


Ducasse:
It is the occasion to get no users in the code.

PR #1749, opened 2025-03-15
Fix: extent when nil returns preferredExtent.

Ducasse:
"may be it means that extent: and extent should not exit and that we should all use preferredExtent and preferredExtent:"


Ducasse:
The comment above is wrong because preferredExtent: is a global one shared by all the instances of a presenter.

PR #1751, opened 2025-03-28
Reduced UIManager api

### LINKED ISSUE #15431: SpInvalidUserInputError MUST not refer to UIManager
```
actForSpec
	"Does nothing on purpose."
	UIManager default
		inform: self messageText.
	self flag: #TODO "In the best world, we should delegate the behaviour to the Spec presenter which originated the error."
```

AlexisCnockaert:
-Called a ```GrowlMorph``` to not have a UIManager call in ```SpInvalidUserInputError```
Fixes https://github.com/pharo-project/pharo/issues/15431 on pharo repo

### DIFF
--- a/src/Spec2-Commander2/SpInvalidUserInputError.extension.st
+++ b/src/Spec2-Commander2/SpInvalidUserInputError.extension.st
@@ -3,9 +3,9 @@ Extension { #name : 'SpInvalidUserInputError' }
 { #category : '*Spec2-Commander2' }
 SpInvalidUserInputError >> actForSpec [
 	"Does nothing on purpose."
- 
-	GrowlMorph new
-		contents: self messageText;
-		openInWorld.
+
+	SpInformDialog new
+		label: self messageText;
+		openDialog.
 	self flag: #TODO "In the best world, we should delegate the behaviour to the Spec presenter which originated the error."
 ]
--- a/src/Spec2-Commander2/SpInvalidUserInputError.extension.st
+++ b/src/Spec2-Commander2/SpInvalidUserInputError.extension.st
@@ -4,8 +4,6 @@ Extension { #name : 'SpInvalidUserInputError' }
 SpInvalidUserInputError >> actForSpec [
 	"Does nothing on purpose."
 
-	SpInformDialog new
-		label: self messageText;
-		openDialog.
+	SpApplication defaultApplication inform: self messageText.
 	self flag: #TODO "In the best world, we should delegate the behaviour to the Spec presenter which originated the error."
 ]
--- a/src/Spec2-Commander2/SpInvalidUserInputError.extension.st
+++ b/src/Spec2-Commander2/SpInvalidUserInputError.extension.st
@@ -3,7 +3,9 @@ Extension { #name : 'SpInvalidUserInputError' }
 { #category : '*Spec2-Commander2' }
 SpInvalidUserInputError >> actForSpec [
 	"Does nothing on purpose."
-	UIManager default
-		inform: self messageText.
+ 
+	GrowlMorph new
+		contents: self messageText;
+		openInWorld.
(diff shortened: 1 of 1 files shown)


src/Spec2-Commander2/SpInvalidUserInputError.extension.st
```diff
@@ -3,7 +3,9 @@ Extension { #name : 'SpInvalidUserInputError' }
 { #category : '*Spec2-Commander2' }
 SpInvalidUserInputError >> actForSpec [
 	"Does nothing on purpose."
-	UIManager default
-		inform: self messageText.
+ 
+	GrowlMorph new
```
demarey: We cannot reference Morphic directly here. 
It should be usable in Gtk also

Ducasse: Indeed!
@AlexisCnockaert

Ducasse: We should use a SpDialog

AlexisCnockaert: Ok nevermind ill do this

src/Spec2-Commander2/SpInvalidUserInputError.extension.st
```diff
@@ -3,7 +3,9 @@ Extension { #name : 'SpInvalidUserInputError' }
 { #category : '*Spec2-Commander2' }
 SpInvalidUserInputError >> actForSpec [
 	"Does nothing on purpose."
-	UIManager default
-		inform: self messageText.
+
+	SpInformDialog new
+		label: self messageText;
+		openDialog.
```
demarey: That part is not easy. A dialog presenter is indeed better but what happens if you run your image headless?
In this case, you do not want to open a dialog but instead log the message on the console.
It is the job done by UIManager.
Currently, in Spec, there is a missing mechanism to deal with headless images.
For now, you could delegate the handling of the inform to the spec application. It already has #inform: message

AlexisCnockaert: Okay I understood, thank you for the explanation

src/Spec2-Commander2/SpInvalidUserInputError.extension.st
```diff
@@ -3,7 +3,7 @@ Extension { #name : 'SpInvalidUserInputError' }
 { #category : '*Spec2-Commander2' }
 SpInvalidUserInputError >> actForSpec [
 	"Does nothing on purpose."
-	UIManager default
-		inform: self messageText.
+
+	SpApplication defaultApplication inform: self messageText.
```
demarey, line 7: Don't you have a reference to the SpApplication instance?
Because you are not sure the application that triggered `SpInvalidUserInputError` is the default or not. It is better to avoid the use of globals.
If you don't have it, you could pass it to SpInvalidUserInputError when creating the error.

Ducasse:
so should I close this PR? or do we let it simply rot?

src/Spec2-Commander2/SpInvalidUserInputError.extension.st
estebanlm: yeah, this is bad. 
also I do not understand where actForSpec is used. That method should not exist at all.

estebanlm (changes requested):

src/Spec2-Commander2/SpInvalidUserInputError.extension.st
tinchodias: I don't see references to SpInteractionError in my Pharo13 but I guerss it could be initialized with the application, couldn't it?

Ducasse:
So I suggest that we close this issue because it is one year old.

Ducasse:
SpInvalidUserInputError does not have reference to application so there is no way we can do something. 
Now SpInvalidUserInputError is not used and its superclass neither.

PR #1752, opened 2025-03-29
add SpFrameLayout to morphic backend

estebanlm:
... and SpTabLayout announce tab selected/unselected

### DIFF
--- a/src/Spec2-Layout/SpFrameLayout.class.st
+++ b/src/Spec2-Layout/SpFrameLayout.class.st
@@ -1,0 +1,114 @@
+"
+A layout with maximally one child that adds a frame around its contnent with optional label
+"
+Class {
+	#name : 'SpFrameLayout',
+	#superclass : 'SpExecutableLayout',
+	#instVars : [
+		'label',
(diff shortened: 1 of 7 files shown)


src/Spec2-Layout/SpFrameLayout.class.st
```diff
@@ -0,0 +1,114 @@
...
+]
+
+{ #category : 'private' }
+SpFrameLayout >> constraintsClass [
+
+	^ SpFrameConstraints
+]
+
+{ #category : 'initialization' }
+SpFrameLayout >> initialize [
```
Ducasse, line 64: Can we removed

src/Spec2-Layout/SpFrameLayout.class.st
```diff
@@ -0,0 +1,114 @@
+"
+A layout with maximally one child that adds a frame around its contnent with optional label
```
Ducasse, line 2: contents

Ducasse:
Would be nice to have tests

PR #1755, opened 2025-04-09
add missing delegate method

estebanlm:


estebanlm:
this is duplicated with #1754

PR #1763, opened 2025-04-25
Update taskbar on open / close windows

PalumboN:
### Before

Screen Recording 2025-04-25 at 17 13 37

In this video I open and close many windows using the shortcuts. However, the taskbar on bottom is not updated _until the cursor is moved_.

### After

Screen Recording 2025-04-25 at 17 20 52

Now it is updated when the windows are opened or closed.

-------

I'm not sure about the implementation, but I was trying other alternatives a lot. Mainly, I wanted to hook some announcement, but they does not work as I expected, and the code comes more complex.


estebanlm:
hello, sorry but this fix is in the bad place. 
If the morphic window is not being updated (which is indeed a bug), then the fix needs to be in morphic level (hence, the `StandardWindow` morph)... under certain circumstances this could be also managed in adapter level (hence SoMorphicWindowAdapter or around). But never in Spec level, which basically does not has to know the existence of a taskbar in the morphic world.

PR #1765, opened 2025-04-30
Making SpCodePresenter reactive to setting change

StevenCostiou:
- When "code show line numbers" settings is changed, an announcement is sent and code presenters automatically update
- The settings is presenter specific, but known by the presenter's application
- The application is responsible to trigger the announcement (in spec, it is not by default but the StPharoApplication from NewTools does)


StevenCostiou:
Unrelated test failing.

PR #1785, opened 2025-06-20
Implemented ```selectAll``` for the tree presenters

AlexisCnockaert:
Improvement of Spec by adding selectAll for trees + 1 test on the selection, more efficient for code usage.
@jecisc its all good normally :)


jecisc (approved):

src/Spec2-Tests/SpAbstractTreePresenterTest.class.st
```diff
@@ -144,6 +144,46 @@ SpAbstractTreePresenterTest >> testPathItemOf [
...
+	presenter selectAll.
+	"in single mode selectAll has no effect"
+	self assertEmpty: presenter selection selectedPaths.
+	presenter beMultipleSelection.
+	presenter selectAll.
+	"but in multiple mode selectAll works" 
+	self
+		assert: presenter selection selectedItems asArray
+		equals:
+			#( 1 10 100 101 102 103 104 105 106 107 108 109 110 11 110 111 112
```
demarey, line 160: would be better to use (1 to: 410) asArray

demarey:
what is the use case for adding this method?
I never needed it and it could be very slow (computing of all children) but you had probably something in mind

AlexisCnockaert:
> what is the use case for adding this method? I never needed it and it could be very slow (computing of all children) but you had probably something in mind
This method addition is linked to this:
https://github.com/pharo-project/pharo/pull/18297#issuecomment-2992205378

jecisc:
It can indeed be slow for huge trees. But on some small ones it can be really useful. 

The usecase here is in DrTest. Alexis is replacing the second pane to go from a list to a tree showing the hierarchy. In order to fix the problem we had when we wanted to run the tests of an abstract class and its subclasses. When we select a package, by default we want to select all tests cases. I doubt we will have a number of test cases big enough to cause performance problems with the select all

PR #1809, opened 2025-08-20
feat: add selectFirst method for SpAbstractTextPresenter

moufort:
Add SelectFirst:ifAbsent: method who select the fist occurence of a String, and do ifAbsent Block when the occurence didn't exist


estebanlm:
I do no understand why this functionality is needed in the presenter. 
Why not just to implement it when, where and how you need it? 

In essence, I think this functionality is of higher level of what a text presenter should be... so, not super convinced of integrate this.

estebanlm:
no, in fact not needed.

PR #1811, opened 2025-08-29
Reduce the number of extensions

guillep:
Many extesions are:

- tests extending existing test classes that can be either integrated in the extended class or put in their own class
- extensions that could be integrated in the extended class but maybe were not because of historical reasons
- extensions that are used in a single place, have only that single use case, and can just be inlined in the caller

This PR is a follow up of https://github.com/pharo-project/pharo/pull/18501 that solves many unjustified extensions using the simple heuristics above:

- move to the extended class
- inline in caller
- put in their own class

WARNING ⚠️  Depends on https://github.com/pharo-project/pharo/pull/18501 and should be integrated after it


estebanlm:
I disagree with the removal of `asValueHolder` (because the replacement lacks the flexibility of it), but the rest is fine.

estebanlm:
tag is because I'm waiting for the other side.

jecisc:
Since this PR is in discussion, I opened another one with a subset of the changes: https://github.com/pharo-spec/Spec/pull/1821

The goal of this PR is to just remove the methods overriding methods from the pharo repository

estebanlm:
so, we should now close this ?

jecisc:
Some things should still be integrated. But a decision needs to be taken for the #asValueHolder. 

I find it easier to us #asValueHolder personally as Esteban

guillep:
Coming back here, please note @jecisc and @estebanlm that this PR does not remove `asValueHolder`. Just remove some usages.

Please tell me if I should close it.

estebanlm:
I am ok with this PR

estebanlm (approved):

PR #1815, opened 2025-09-03
Fix Active presenter of a Notebook is not available in the #whenSelectedTabChangedDo:

estebanlm:


src/Spec2-Core/SpNotebookPresenter.class.st
```diff
@@ -136,7 +131,12 @@ SpNotebookPresenter >> resetAllPageContents [
 SpNotebookPresenter >> selectPage: aPage [
 	"Selects the page `aPage` from the notebook, switching to show it."
 
-	aPage = self selectedPage ifTrue: [ ^ self ].
+	"(aPage isNotNil 
+		and: [ aPage = self selectedPage 
+		and: [ aPage activePresenter isNotNil ] ]) 
+		ifTrue: [ ^ self ]."
```
demarey: debug code to be removed?

PR #1819, opened 2025-09-11
move  to Clap

demarey:


estebanlm (approved):
I would rename ClapPharoApplication since is confusing (it seems like an Spec application, and is a Clap application and meh...) but otherwise is fine ;)

Issue #1818, opened 2025-09-04
Bug: A SpBoxLayout with a border gets extra width/height when adding presenters


koendehondt:
### Step 1

Consider this code:
```smalltalk
	application := SpApplication new
		addStyleSheetFromString: '.application [
			.tab [ Container { #borderWidth: 3, #borderColor: #FFFFFF } ] ]'.
	mainPresenter := SpPresenter newApplication: application.
	iconLayout := SpBoxLayout newLeftToRight
			add: (mainPresenter newImage image: (mainPresenter iconNamed: 'box')) expand: false;
			yourself.
	icon := (SpPresenter newApplication: application)
		layout: iconLayout;
		addStyle: 'tab';
		yourself.
	iconBar := (SpPresenter newApplication: application)
		layout: (SpBoxLayout newLeftToRight
			spacing: 10;
			add: icon expand: false;
			yourself);
		yourself.
	mainLayout := SpBoxLayout newTopToBottom
		vAlignCenter;
		hAlignCenter;
		yourself.
	mainLayout add: iconBar expand: false.
	mainPresenter layout: mainLayout; open.
```
It creates an `iconBar ` presenter with a `SpBoxLayout` that holds 1 `icon` presenter. The latter holds an image. The`icon` presenter has a white border of 3 pixels to show the issue clearly later on.

The code snippet opens this window:

<img width="255" height="140" alt="Image" src="https://github.com/user-attachments/assets/43a17c63-4972-4478-a56a-65dc2cd6e8ac" />

All good. The 3-pixel white border encloses the image as expected.

### Step 2

Change the code to put 2 images in the `iconLayout`:

```smalltalk
iconLayout := SpBoxLayout newLeftToRight
			add: (mainPresenter newImage image: (mainPresenter iconNamed: 'box')) expand: false;
			add: (mainPresenter newImage image: (mainPresenter iconNamed: 'box')) expand: false;
			yourself.
```

and run the code again. Now this window opens:

<img width="256" height="141" alt="Image" src="https://github.com/user-attachments/assets/10fdb8b5-44ae-4534-812d-810a4a32b49c" />

The white border is now wider than the space taken by the images. The extra space inside the border is 6 pixels.

### Step 3

Change the code again, now to put 3 images in the `iconLayout`:

```smalltalk
iconLayout := SpBoxLayout newLeftToRight
			add: (mainPresenter newImage image: (mainPresenter iconNamed: 'box')) expand: false;
[...]

Issue #1525, opened 2024-02-24
The box layout algorithm does not handle spacing of a SpBoxLayout correctly


koendehondt:
Examples from a recent Pharo 12 image.

### Box layout without spacing

Consider this code:

```smalltalk
	| mainLayout mainPresenter iconAndLabel tabBar application tab |
	application := SpApplication new
		addStyleSheetFromString: '.application [
			.tab [ Container { #borderWidth: 1, #borderColor: #FFFFFF } ] ]'.
	mainPresenter := SpPresenter newApplication: application.
	iconAndLabel := (SpPresenter newApplication: application)
		layout: (SpBoxLayout newLeftToRight
			vAlignCenter;
			add: (mainPresenter newImage image: (mainPresenter iconNamed: 'box')) expand: false;
			add: (mainPresenter newLabel label: 'Some label') expand: true;
			yourself);
		yourself.
	tab := (SpPresenter newApplication: application)
		addStyle: 'tab';
		layout: (SpBoxLayout newTopToBottom
			add: iconAndLabel expand: true;
			yourself);
		yourself.
	tabBar := (SpPresenter newApplication: application)
		layout: (SpBoxLayout newLeftToRight
			spacing: 10;
			add: tab expand: false;
			yourself);
		yourself. 
	mainLayout := SpBoxLayout newTopToBottom
		add: tabBar height: 40;
		yourself.
	mainPresenter layout: mainLayout; open
```

This is the foundation of displaying a horizontal tab bar, where a tab consists of an image and a label. The image and the label have a horizontal layout. A tab is styled with a white border to show the bug clearly later on.

The code above opens this window:

<img width="400" alt="Screenshot 2024-02-24 at 15 47 13" src="https://github.com/pharo-spec/Spec/assets/753271/774344d2-d53f-4251-b209-fcb3abe15bff">

Everything is displayed as expected: there is no white space between the image and the label. There is no space between the image and the white border, nor between the label and the white border. All good.

### Box layout with spacing

Now suppose that some spacing is needed between the image and the label. Given that the image and the label are presenters in a `SpBoxLayout`, the new requirement can be implemented by using `#spacing:`. Here is the code again, with `spacing: 20;` added (the 8th line in the code):

```smalltalk
	| mainLayout mainPresenter iconAndLabel tabBar application tab |
	application := SpApplication new
		addStyleSheetFromString: '.application [
[...]

PR #1820, opened 2025-09-12
Fix SpMorphicBoxAdapter>>#verifyBoxExtentOf:withChild: to only add border width/height when adding the first child morph instead of adding border width/height for every child morph

koendehondt:
This PR fixes bug 1 as reported by  https://github.com/pharo-spec/Spec/issues/1818

In the screenshots, we see a border of 3 pixels. We use presenters with a fixed size, i.e. an image of 16 by 16 pixels. See the added unit tests.

### Before situation

For a layout with one presenter, the calculation of the extent is ok:

Image

For a horizontal layout with more than one presenter, the calculation of the extent is wrong, as described in https://github.com/pharo-spec/Spec/issues/1818. Here is the situation when the layout holds three presenters:

Image

For a verical layout with more than one presenter, the calculation of the extent is also wrong. Here is the situation when the layout holds three presenters:

Screenshot 2025-09-12 at 08 36 57

### After situation

This PR fixes the bug so that we have these results:

Screenshot 2025-09-12 at 08 37 35

Screenshot 2025-09-12 at 08 37 42

Screenshot 2025-09-12 at 08 37 58

All Spec tests have been run. 3 tests failed, but they already failed before the changes in this PR:

Screenshot 2025-09-12 at 09 08 30

### DIFF
--- a/src/Spec2-Adapters-Morphic/SpMorphicBoxAdapter.class.st
+++ b/src/Spec2-Adapters-Morphic/SpMorphicBoxAdapter.class.st
@@ -332,8 +332,12 @@ SpMorphicBoxAdapter >> verifyBoxExtent [
 SpMorphicBoxAdapter >> verifyBoxExtentOf: aPanel withChild: childMorph [
 	| width height newPanelWidth newPanelHeight |
 
-	width := childMorph width + (widget borderWidth * 2).
-	height := childMorph height + (widget borderWidth * 2).
+	width := childMorph width.
+	height := childMorph height.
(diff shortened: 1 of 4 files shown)


koendehondt:
@estebanlm Please take a look at this PR, as I would like to fix a related bug in the same code (https://github.com/pharo-spec/Spec/issues/1525).

koendehondt:
@estebanlm, please review and merge if the changes are acceptable. I need this fix to fix other bugs. Thanks.

PR #1822, opened 2025-09-26
Add ```ifNone:``` to the detect in method

AlexisCnockaert:
This change is made because we call this method to update the ```Iceberg browser```

### DIFF
--- a/src/Spec2-Core/SpNotebookPresenter.class.st
+++ b/src/Spec2-Core/SpNotebookPresenter.class.st
@@ -72,7 +72,7 @@ SpNotebookPresenter >> pageAt: index [
 SpNotebookPresenter >> pageNamed: aTitle [
 	"Answer the receiver's <SpNotebookPage> with aTitle as name. Raise a <NotFound> exception if not present"
 
-	^ self pages detect: [ : page | page title = aTitle ]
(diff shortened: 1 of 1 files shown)


src/Spec2-Core/SpNotebookPresenter.class.st
```diff
@@ -72,7 +72,7 @@ SpNotebookPresenter >> pageAt: index [
 SpNotebookPresenter >> pageNamed: aTitle [
 	"Answer the receiver's <SpNotebookPage> with aTitle as name. Raise a <NotFound> exception if not present"
 
-	^ self pages detect: [ : page | page title = aTitle ]
+	^ self pages detect: [ :page | page title = aTitle ] ifNone: [ ^ nil ]
```
demarey, line 75: why returning nil instead of getting a NotFound exception?

AlexisCnockaert: Because I need it to return nil for this piece of code ``` (repositoryNotebook pageNamed: group label) ifNil: [ code ]```

AlexisCnockaert: My bad I can handle it in an other way, I will close this

PR #1831, opened 2025-10-26
Making sure that cm2 commands with their spec extensions can talk to the application

Ducasse:


src/Spec2-Commander2/CmCommandAborted.extension.st
```diff
@@ -2,5 +2,12 @@ Extension { #name : 'CmCommandAborted' }
 
 { #category : '*Spec2-Commander2' }
 CmCommandAborted >> actForSpec [
-	UIManager default inform: ('"{1}" command was aborted because: {2}' format: { self command name . self reason })
+	application inform: ('"{1}" command was aborted because: {2}' format: { self command name . self reason })
```
demarey, line 5: What happens if application is not set? i.e. we should either not inform or find another way to inform

Issue #1830, opened 2025-10-24
broken paginator in inspector because of some missing thing in SpBoxLayout changes


estebanlm:
https://github.com/pharo-spec/Spec/pull/1813

this PR introduced a problem in the calculus of box size, at least in the case of the paginator.

estebanlm:
cc @koendehondt :)

koendehondt:
@estebanlm and @Ducasse 

Next week, I will attend the Pharo Sprint. That is a good time to investigate the issue. Then we can also talk about it.
But it would be helpful to have a description of the paginator's expected behaviour. Could you add it to this ticket?
With that description, I can take a look at the issue before the Pharo Sprint.

estebanlm:
Hello, well... this is easy. I expect it to be visible :) 

<img width="1053" height="808" alt="Image" src="https://github.com/user-attachments/assets/ebd4becc-4aac-4a25-b621-c620a9a5a993" />

next sprint is highly unlikely I will manage to go as this is school holidays, and I will most probably need to work from home. 
I will try to be there at least a bit, but no promises :(

koendehondt:
The PR I submitted was rejected by @estebanlm. I re-investigated the issue reported in #1735 and "fixed" in #1813.

This ticket proves that the issue reported in #1735 was not fixed by #1813. I know why. I'd like to explain below.

### Inspector

It all starts in `StInspectorPresenter>>#initializePresenters`. It creates a `SpMillerPaginatorPresenter`.
```smalltalk
initializePresenters

	millerList := self instantiate: SpMillerPaginatorPresenter.
	millerList millerListPresenter
		spacing: self spacing;
		presenterBlock: [ :aModel | self newInspectorFor: aModel ];
		setRootModel: self model;
		whenColumnsWillChangeDo: [ self keepLastPageSelectedTabName ]
```

`SpMillerPaginatorPresenter>>#initializePresenters` creates a `SpMillerColumnPresenter` and, more interestingly, an empty `SpBoxLayout`, which is added to the main layout with `add: ... expand: false`. The presenter keeps the empty box layout in an instance variable `paginatorContainerLayout`. Here is the implementation:
```smaltalk
initializePresenters

	self layout: (SpBoxLayout newTopToBottom
		add: (millerList := self instantiate: SpMillerColumnPresenter);
		add: ((paginatorContainerLayout := SpBoxLayout newLeftToRight)
				hAlignCenter; 
				yourself)
			expand: false;
		yourself).

[...]

PR #1838, opened 2025-11-01
Make the height of the paginator in a SpMillerPaginatorPresenter 20 pixels high explicitly

koendehondt:
Fix for #1830.

Here is the result after the fix. See inside the green circle.

Screenshot 2025-11-01 at 12 25 08

All Spec tests have been run:

Screenshot 2025-11-01 at 12 18 12

### DIFF
--- a/src/Spec2-Core/SpMillerPaginatorPresenter.class.st
+++ b/src/Spec2-Core/SpMillerPaginatorPresenter.class.st
@@ -41,7 +41,7 @@ SpMillerPaginatorPresenter >> initializePresenters [
 		add: ((paginatorContainerLayout := SpBoxLayout newLeftToRight)
 				hAlignCenter; 
 				yourself)
-			expand: false;
+			height: 20;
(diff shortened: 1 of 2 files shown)


estebanlm (changes requested):
I am sorry, I do not like this change. It is not fixing a problem (which seems to be calculating the height of a non-expanded row in a box layout) and instead is assigning a fixed (hardcoded) height to it which is not the case. 
So, we need to determine why the paginator is not getting the right height and that may be for two reasons: 
1. the box layout fails somehow to calculate it.
2. is the paginator component who is not declaring correctly its requested height (but this was working before your changes to the layout, so no idea :) ).

koendehondt:
No worries. I will investigate it further. It is hard to debug because box layouts are used everywhere.

koendehondt:
This PR is not relevant anymore after @estebanlm's review comments. It has been replaced by #1843.

PR #1844, opened 2025-11-19
Make #inform: handle non interactive images

jecisc:


jecisc:
Steph agrees with the change, I'll merge

PR #1856, opened 2025-11-27
add priority to SpActions

estebanlm:
... so they can be sorted when needed, other than alphabetically by name.
Also enhanced some class comments..


estebanlm:
errors unrelated

PR #1857, opened 2025-11-28
enhance some class comments

estebanlm:


Ducasse:
@estebanlm I read the comments and they are better :)
Now just a question what people should use? Should they use the easy family and in the future we can deprecate the others?
This message was not clear to me. 
For example for the book should we migrate all the code to use easy?
Is it done automatically when we use newList?

estebanlm:
yes, in the future we will deprecate the others (not yet, however :P).

in fact, I will eventually changing the code to answer a 
`SpEasyListViewPresenter` instead `SpListPresenter when sending 
`newList` (not yet, however) :P

PR #1860, opened 2025-12-02
making list family behave a bit better when disabling/enabling activation/selection

estebanlm:
I took the solution from Pablo's https://github.com/pharo-project/pharo/pull/18902 that makes the same with announcements.

Then I was thinking this would make also sense to bind this to a process but in the case of, e.g. Spec-Gtk is completely useless (as each callback/event happens in a separated thread)...


estebanlm:
ok, I do not think this is relevant anymore (this was fixed upstream)

PR #1864, opened 2025-12-15
Revert "Revert "New rule: Presenter should not redefine `new`""

jecisc:
Reverts pharo-spec/Spec#1863

In order to merge this PR we need to load Spec after the model of rules in Pharo


estebanlm:
so... ?

tinchodias:
What is this?

Ducasse:
It was about a rule that make sure that people do not use new for Presenter but only newApplication: but it did not work.

estebanlm:
yeah... the idea is that no presenter can be created ourside its POM (Presenter Object Model). 
The problem is that sometimes you need to first create a presenter, and after add it to a POM, and that is complicated to change. 
but other times people just create them just because they can... and that's bad.
I was thinking to create a explicit way like adding some `newDetached` for example... and a validation when things will be displayed... that at least would make the problem explicit.

PR #1865, opened 2026-01-12
Extension cleanup

guillep:
Move extensions to proper places:
 - to the defining classes if possible
 - to the user if (and very clear) there is only one
 - close to other similar extensions if possible

Related to https://github.com/pharo-project/pharo/pull/19150


Ducasse:
I imagine that there is a cycle between spec and morphic and this is why the tests are breaking.

PR #1874, opened 2026-03-04
SpTextInputFieldAdapterTest: talk to adapter instead of widget [P14]

tinchodias:
The backend is a parameter of this suite, then tests should send messages to the adapter.

Porting https://github.com/pharo-spec/Spec/pull/1873


Ducasse:
I like the change! Less breaking layers!

PR #1878, opened 2026-03-07
Moving performTest logic to setUp!

Ducasse:


tinchodias:
This PR looks good, but it seems `SpMorphicAdapterTest` has no tests and no subclasses. Is this right? or is this class a candidate for removal?

I downloaded a fresh Pharo 14, I added Spec repo and pulled Pharo14 branch: 
Screenshot 2026-03-13 at 12 54 51

Ducasse:
I do not know just fixing all the performTest from the image.

Ducasse:
If the class is useless we should remove it.

PR #1880, opened 2026-03-13
Improve SpGridLayoutAdapterTest [P13]

tinchodias:
- avoid using newMorph
- use only adapter API instead of Morphic API

Matches #1879 for P13


Ducasse:
Hi martin

Do we apply this to P14?
Or both P13 and P14?
S.

tinchodias:
Hi Stef, yes, you merged #1879 which was the same change but for the Pharo14 branch

PR #1888, opened 2026-04-16
modify SpFilteringSelectableListPresenter

estebanlm:
because the old version was somehow buggy


estebanlm:
now errors are not related

PR #1891, opened 2026-04-20
introduce SpCodeEditorPresenter

estebanlm:
to handle highlevel code edition

### DIFF
--- a/src/Spec2-Adapters-Morphic/SpAbstractMorphicAdapter.class.st
+++ b/src/Spec2-Adapters-Morphic/SpAbstractMorphicAdapter.class.st
@@ -465,32 +465,31 @@ SpAbstractMorphicAdapter >> keyDown: keyCode shift: shiftBoolean meta: metaBoole
 
 { #category : 'emulating' }
 SpAbstractMorphicAdapter >> keyEvent: type code: keyCode shift: shiftBoolean meta: metaBoolean control: controlBoolean option: optionBoolean [
+
 	| evt modifiers |
-	
 	modifiers := 0.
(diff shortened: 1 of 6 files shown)


estebanlm:
errors not related

Ducasse:
Cool we will use it. We will migrate all the refactoring commands to be able to expose the refacforings and more to Spec.

Ducasse:
@AlexisCnockaert

Ducasse:
@estebanlm do we get menu builder with this one also? I imagine. Because this is the hook we need to be able to do  node selection. BTW why this editor does not have an interaction model?

estebanlm:
I already added it to `StMethodBrowser` :)
I was thinking there are things from the method browser that can be merged in the editor, instead... but who knows. Is something to analyse... maybe we need a more specific children "SpMethodEditorPresenter ;)

estebanlm:
> @estebanlm do we get menu builder with this one also? I imagine. Because this is the hook we need to be able to do node selection. 

It will, the editor is not finished, I want to check first which methods are worthy to be delegated and which not. 

> BTW why this editor does not have an interaction model?

it has `beFor*`, I just didn't exposed directly `interactionModel:`. 

I have doubts about  some of those methods, I need to think. The general idea is that the "common" things to use should be exposed and the more "low level" should be accessible through `codePresenter` accessor... is maybe a bad idea, I am not sure... this is still WIP!!!! ;)

Ducasse:
Ok I understand. The separation model / editor was not that clear to me either.

Ducasse:
maybe we need a more specific children "SpMethodEditorPresenter ;)
Yes and one for class definition too.

PR #1892, opened 2026-04-20
Introduce refactoring commands for Spec

AlexisCnockaert:
This is the start of the migration of refactoring commands, used now for the method browser. 

For now I implemented : 
-`Rename Class`
-`Extract Temp`
-`Convert temp to inst var`

We could use these for other places that deal with the code presenter (Playground,...) but we will need an unified API.

I also added test for the commands to check if the navigation through AST is well plugged to drivers.
Created a mock context for the tests, also need a special PR for the driver in pharo repo


AlexisCnockaert:
Failing test for selector not implemented is normal since this is introduced in this PR

MarcusDenker:
I think the PR has a code loading problem: https://github.com/pharo-project/pharo/issues/19590

AlexisCnockaert:
@MarcusDenker I will try to fix this, btw could you look to merge https://github.com/pharo-spec/NewTools/pull/1403 please, because New Tools CI is failing too

MarcusDenker:
> @MarcusDenker I will try to fix this, btw could you look to merge [pharo-spec/NewTools#1403](https://github.com/pharo-spec/NewTools/pull/1403) please, because New Tools CI is failing too

merged!

PR #1899, opened 2026-04-30
Introduce message: (with deprecation on SpInformUserDialog) to be com…

Ducasse:
Introduce message: (with deprecation on SpInformUserDialog) to be compliant with the API of Sp**Dialog


estebanlm:
weird that I missed to implement it myself but well... ;) 
(maybe I just override it with an image in a previous version... I need to stop jumping from image to image :P)

estebanlm (approved):

PR #1901, opened 2026-04-30
Change deprecation to make those tests use the right API

NathanMalenge:
Change deprecation to make those tests use the right API


estebanlm (approved):
nice, I introduce that to be more coherent with the rest of the system, I think it fits better (also it introduces a polymorphic way of create presenters, which is  useful in certain cases)

PR #1918, opened 2026-06-10
add expansion events to old trees 

estebanlm:
because they are still in use so...

also introduce a deprecated method #isExpand since I am permanently problems because they are not unified.


estebanlm:
error not related

PR #1923, opened 2026-06-25
Fix code presenter Browse Implementors

AlexisCnockaert:
- So now it calls the new method browser, and we clean up the callsites

### DIFF
--- a/src/Spec2-Code/SpCodePresenter.class.st
+++ b/src/Spec2-Code/SpCodePresenter.class.st
@@ -272,16 +272,17 @@ SpCodePresenter >> doBrowseHierarchy [
 
 { #category : 'commands' }
 SpCodePresenter >> doBrowseImplementors [
+
 	| variableOrClassName variable |
-	
 	variableOrClassName := self selectedSelector ifNil: [ ^ nil ].
 	variable := self lookupEnvironment lookupVar: variableOrClassName.
-	
+
 	"For global and class variables, implementors-of browses the class of the value of the variable, 
 	for classes this means it browses the class"
-	(variable isNotNil and: [variable isGlobalVariable or: [variable isClassVariable]]) 
+	(variable isNotNil and: [ variable isGlobalVariable or: [ variable isClassVariable ] ])
 		ifTrue: [ self systemNavigation browse: variable value ]
-		ifFalse: [ self systemNavigation browseAllImplementorsOf: variableOrClassName ]
+		ifFalse: [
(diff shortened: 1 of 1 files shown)


guillep:
Hi, this fix is wrong.

Tests are breaking because they were expecting to pass through system navigation and this is now hardcoding Smalltalk, tools, and other stuff again.

I'll revert it.

src/Spec2-Code/SpCodePresenter.class.st
```diff
@@ -272,16 +272,17 @@ SpCodePresenter >> doBrowseHierarchy [
...
-	
+
 	"For global and class variables, implementors-of browses the class of the value of the variable, 
 	for classes this means it browses the class"
-	(variable isNotNil and: [variable isGlobalVariable or: [variable isClassVariable]]) 
+	(variable isNotNil and: [ variable isGlobalVariable or: [ variable isClassVariable ] ])
 		ifTrue: [ self systemNavigation browse: variable value ]
-		ifFalse: [ self systemNavigation browseAllImplementorsOf: variableOrClassName ]
+		ifFalse: [
+		(Smalltalk tools toolNamed: #messageList) browseImplementorsOfAll: { variableOrClassName } from: self interactionModel behavior ]
```
guillep, line 285: The right fix should be to pass though system navigation here, as it was before.

If we need to add an argument for the origin class, that should be extended there too.
And maybe we need to support the old method too for some time as a compatibility measure.

guillep: This is the cause of the failing tests:

- SpCodePresenterTest testDoBrowseSenders
- SpCodePresenterTest testDoBrowseSendersOfClassNameShouldNotFindIfNotInEnvironment
- SpCodePresenterTest testDoBrowseSendersOfClassNameWhenInPlaygroundButNotAvailable
- SpCodePresenterTest testDoBrowseSendersOfClassPoolVariableShouldSelectClassPoolVariableUsersWhenInDebuggerButNotAvailable
- SpCodePresenterTest testDoBrowseSendersOfClassVariableShouldSelectClassVariableUsersWhenInDebuggerButNotAvailable
- SpCodePresenterTest testDoBrowseSendersOfClassVariableShouldSelectClassVariableUsersWhenInStaticContextButNotAvailable

AlexisCnockaert:
Hi, I understand why, I did this because the new method browser works with tabs now. Therefore if someone want to browse implementors in a method browser it will not open as a tab but as a new window since it calls the system navigation.
But looking at the trade-offs it is better to revert this.

guillep:
My point is not that the feature is bad! My point is that we should redo it, just being careful of respecting the architecture and keeping tests green!! :)

PR #1926, opened 2026-06-25
add api to access background color at adapter level to avoid breaking spec backend layer

### LINKED ISSUE #1925: The test `testBackgroundColorChangesColor` is Morphic specific
The test is Morphic specific accessing API of Morphic directly

badetitou:
fix #1925


tinchodias (approved):
It's right. I did several fixes of this kind: the adapter tests shouldn't talk to the widget directly (else, it is assuming the widget is a Morph)

PR #1929, opened 2026-06-26
Open browsers on literal and not their names

guillep:
When opening a method browser, the reference is to a literal and not to its name or a string representation of it.

Otherwise, it does not allow proper refresh.

This is causing right now that refreshing removes elements from the method browser


guillep:
Related to https://github.com/pharo-project/pharo/pull/19829

Issue #1933, opened 2026-07-08
How to modify `SpAbstractAdapterTest` to support async backend


badetitou:
I investigate how to make tests of SpecToplo working, and I encounter an issue with many of the tests here.
Particularly with the tests when selecting item of a list (or tree, table, treeTable).

The selection is performed in the UI process and not in the test one; thus, often the `assert:` is call before the UI had time to perform the select action.

I have two proposals:

1. add `backendForTest waitUntilUIRedrawed.` before every `assert:` to ensure that we are in a expected context
2. implement another `assert:` that includes the `waitUntilUIRedrawed` and would take the condition under evaluation as a block

A better solution would be to force some kind of using the same process for test in `SpToploBackendForTest` but I have no clue where I should start to do something like this, probably with `runTest:`

badetitou:
One solution could be to add in https://github.com/pharo-spec/Spec/blob/724ca7484c00cda961277de9367f0e798f64c8e4/src/Spec2-Backend-Tests/SpAbstractAdapterTest.class.st#L32 a call to 

```st
	backendForTest waitUntilUIRedrawed.
```

So in many test that ends like this

```st
testSelectPresenterIndexSetsSelectedIndexInWidget
	presenter selectIndex: 1.
	self assert: (self adapter selectedIndexes includes: 1)
```

the `self adapter` calls in the `assert` will ensure everything is already performed in the UI part

Any opinion @tinchodias?

tinchodias:
I agree `waitUntilUIRedrawed` is needed and like your idea of sending it in from a "convenience" method in the Test hierarchy. But I'd like an alternative to sending it implicitly in `adapter`... It looks like an accessor and it'd be doing a wait for other Process, and this sounds too relevant to me to do it implicitly. Maybe `adapterAfterWait`, `syncAdapter", `waitAdapter`? other?

badetitou:
I understand. I'd like to avoid updating all tests, that was the point. 
But I can make this effort to make the tests pass :)

badetitou:
It is also only in the context of testing. So it might be documented. 

I think one underlying question is how far we don't want to update tests to be as close as possible as original ones? 
I'll comply with this choice

tinchodias:
> It is also only in the context of testing. So it might be documented.

Okay, you convinced me, no problem on my side!

PR #1934, opened 2026-07-08
in test, ensure we wait interaction to finish before asserting the result

badetitou:
This PR is link to the issue here https://github.com/pharo-spec/Spec/issues/1933 even in the cases the proposed solution (second comment) is not accepted


Ducasse:
@estebanlm can you have a look? tx

estebanlm (approved):
sounds like correct :)

PR #1935, opened 2026-07-08
make the accessor to adapter also checking the UI state

### LINKED ISSUE #1933: How to modify `SpAbstractAdapterTest` to support async backend

badetitou:
fix #1933 

Please look at the issue discussion


Ducasse:
@estebanlm can you have a look? tx

PR #1939, opened 2026-08-31
Callsite fix for Implementors in code presenter

AlexisCnockaert:
So the `SpCodePresenter` calls the tool directly, allowing it to open in the method browser tabs

### DIFF
--- a/src/Spec2-Code/SpCodePresenter.class.st
+++ b/src/Spec2-Code/SpCodePresenter.class.st
@@ -280,8 +280,8 @@ SpCodePresenter >> doBrowseImplementors [
 	"For global and class variables, implementors-of browses the class of the value of the variable, 
 	for classes this means it browses the class"
 	(variable isNotNil and: [ variable isGlobalVariable or: [ variable isClassVariable ] ])
-		ifTrue: [ self systemNavigation browse: variable value ]
-		ifFalse: [ StMethodBrowser browseImplementorsOfAll: { variableOrClassName } from: self behavior ] 
+		ifTrue: [ variable value browse ]
+		ifFalse: [ StMethodBrowser browseImplementorsOfAll: { variableOrClassName } from: self behavior ]
 ]
 
 { #category : 'commands' }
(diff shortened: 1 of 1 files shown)


estebanlm (changes requested):
this is the same problem as with doBrowseSenders. You should not call this directly but through SystemNavigation :)

Ducasse:
We should not use browse and other from system navigation. This architecture is a crap. 
We should stop to use it and I wanted to discuss this during the meeting.

Ducasse:
The StApplication should be enhanced and used and systemNavigation should not be used anymore. 
All the logic behind using Smalltalk tools is lame.

estebanlm:
let's not change horses in the middle of the ride. 
the fact that SystemNavigation is not a satisfactory answer (I agree) does not means not using just like that it is better.  

in these cases, 
- if you use `Smalltalk tools` I need to register the tool to use globaly and I cannot adapt my application without changing the whole system behavior and this is obviously bad.
- Using directly the presenter (`StMethodBrowser`), without intermediary, means the user is forced to use *that presenter* (and no other) to show a method list to the user. 

with `SystemNavigation` (not great, but there), I just need to give a new instance of it and everything can be adapted as desired. E.g. I can make things like this (a popover, not a dialog), without touching the code presenter at all: 

image
 
without using `SystemNavigation` I need to 
- subclass SpCodePresenter and override the methods to have my desired behavior, or...
- re-do my own set of commands that will call *other methods* to get the desired behavior
- ... (there has to be other ways, but none of them is easier than just subclassing `SystemNavigation` and give it to my presenter).

So, I will still ask to use `SystemNavigation`, even if not ideal. 
- This will avoid mix of designs for the same purpose on the system (`SystemNavigation` is already there, and used) 
- this will also concentrate the refactoring place when we decide how to change it.

And then yes, I would say we need to sit and design a solution better than what we have now :)

estebanlm:
Thinking in loud, the long term solution would be to decouple the command execution from the presenters, and let users define their own.  
But I still think than *for the moment* we need to keep using `SystemNavigation`.

PR #1941, opened 2026-09-09
use SystemNavigation for browsing senders

estebanlm:
even if we need to remove SystemNavigation at a point, right now the consequences of not using it *right now* are far worse.
also, sneak some missing methods.


estebanlm:
errors not related

PR #1942, opened 2026-09-09
senders should go through system navigation

guillep:
Fix SpCodePresenter tests

Navigation should go through the presenter's system navigation and not short-circuit directly using Smalltalk tools.

Two other companion PRs arriving in newtools and Pharo


estebanlm:
lol, I was fixing the same :)

PR #1943, opened 2026-09-09
We need to correctly unmark the dirtyness.

guillep:
This is a workaround for Pharo 14.

Once compilation is done, the method installer will announce it. And the method browser will catch it and see it was dirty and launch a popup. And this happens before the `submit` callback returns! So this is not clear where to do it.
	
But this is not so easy because here we are deep in the presenter hierarchy. Our owner owner may be a SpCodeEditor or not!
	
This needs to be redesigned somehow...

### DIFF
--- a/src/Spec2-Code/SpCodeInteractionModel.class.st
+++ b/src/Spec2-Code/SpCodeInteractionModel.class.st
@@ -58,7 +58,23 @@ SpCodeInteractionModel >> compiler [
 
 { #category : 'interactive error protocol' }
 SpCodeInteractionModel >> correctFrom: start to: stop with: aString [ 
-	self owner adapter widget correctFrom: start to: stop with: aString 
+	"this line of code marks the underlying widget as dirty.
+	But... see below"
+	self owner adapter widget correctFrom: start to: stop with: aString.
+	
+	"We need to unmark the dirtyness here.
+
+	Otherwise, once compilation is done, the method installer will announce it.
+	And the method browser will catch it and see it was dirty and launch a popup.
+	And this happens before the `submit` callback returns!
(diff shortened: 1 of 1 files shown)


estebanlm:
I'm sorry I do not understand, in which conditions is the "popup" showing up ? I cannot see what is this workaround trying to fix :(

guillep:
Sorry, I did not explain!!

# The issue and how to reproduce

The issue is as follows:

1. you're on the method browser
2. you edit the method, using a temp `xxx` that does not exist

imagen

4. accept it => now you have the typical popup "I don't know what xxx is"

imagen

6. select "define as temp"

=> **The bug**, you have this strange popup saying that changes have not been saved

imagen

# The "cause"

- first, the compiler fixes the code when we select "define as temp", marking the presenter as dirty
- after compilation is done and the method is installed, the method installer announces a new method is in the block.
- the method browser listens to that announcement and says "oh, there is this new method, that is the same I'm editing, but the source presenter is marked as dirty, let's ask the user what to do"

I know this PR is not the final correct solution but to me it's not so easy.
The point is that when the announcement happens, the presenter is still in a catatonic state: the compiler has not returned yet, and did not give the power to the presenter to undirt the presenter.

First there are some design decisions to visit:
 - Why does the "compiler edition" mark it as dirty? Should it?
 - Is it the installer that should announce?
 - Maybe it's the why the announcer is handled?

Notice that "unmarkDirty" before compilation does not work, because the popup "define temp" will mark it as dirty again.

guillep:
Superseeded by https://github.com/pharo-spec/NewTools/pull/1527 which is much nicer

Issue #1291, opened 2022-06-29
Feature: Allow second pane to keep size when owner resized


JanBliznicenko:
What I need is to have right/bottom-side toolbar with fixed size considering outside source of resize (like resizing window), but still allow manual resizing by using the splitter. It would also allow to have containers like this: https://youtu.be/pO4J6JflZeg?t=10

Currently, if two panes with a separator are used (for example by SpPanedLayout) and I set fixed position of separator (ie. not percentage) and then their owner is resized (for example by resizing the window or some other outer pane), the first (left/top) pane keeps its size and the second (right/bottom) one is being resized to compensate for the new owner size.

I would like to be able to set position not from the left/top (size of the left/top pane), but position from the right/bottom (size of the right/bottom pane) and once I do, when owner/window is resized, the left/top pane would be resized, not the right/bottom one - it would keep its size like a proper toolbar, yet still allow to be resized using separator.

PR #1945, opened 2026-09-11
PanedLayout now preserves offset when moved, and allows offset from the right

### LINKED ISSUE #1292: Using separator switches paned layout to proportional
Currently, when I use SpPanedLayout and set position as fixed size (not percentage), it keeps the first pane size when resizing window/owner and only the second pane is being resized when owner/window is resized. That is good - how is it intended. For example, when I need left toolbar with fixed width even after window is resized, yet still allow it to be resized manually using separator.

However, once I use the separator (and resize the first pane with fixed size), the layout switches from fixed size of the first pane to proportional (percentual) sizes, so after using separator, the first pane now starts moving when owner/window is resized and no longer acts as a side toolbar.

The separator should honor the original layout (ie. just change the value of fixed size) and not switch to proportional one.

JanBliznicenko:
2 PanedLayout improvements:

* Fixes: https://github.com/pharo-spec/Spec/issues/1292
* Adds: https://github.com/pharo-spec/Spec/issues/1291

### DIFF
--- a/src/Spec2-Adapters-Morphic/LayoutPolicy.extension.st
+++ b/src/Spec2-Adapters-Morphic/LayoutPolicy.extension.st
@@ -1,0 +1,6 @@
+Extension { #name : 'LayoutPolicy' }
+
+{ #category : '*Spec2-Adapters-Morphic' }
+LayoutPolicy >> isFromEnd [
+	^ false
+]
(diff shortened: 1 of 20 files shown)


JanBliznicenko:
Either the failed tests are unrelated or I do not understand what they actually check :)

estebanlm:
I will take a look but 
a) this very complicated to achieve in other backends and this is something that needs to be taken into account before implementing it.
b) in any case, this will not be available in Pharo14 because is a good example of a functionality that takes time to tweak and get it right.

JanBliznicenko:
I am closing this PR, replacing it by one rebased onto Pharo 15:
https://github.com/pharo-spec/Spec/pull/1946

